PR #9, opened 2013-08-05
String(),SaveToFile() and Get/Set Comment Support

achun:
adds the following features

type Comment struct {
    Multiline []string
    EndOfLine string
}

GetComment(key string) Comment
SetComment(key string, endofline string)
SetComments(key string, multiline ...string)
String() string
SaveToFile() error


drauschenbach:
I've arrived here looking for `SaveToFile()` support, and maybe I can help explain the use case.

I use TOML in the first place because it allows comments. This allows me to ship software packages with a fully self-documented config file. These default config files are long, because they document each and every key, but, initially, they are essentially empty because every single key is disabled. This is common practice elsewhere, where I picked up the habit, such as in Apache2's `httpd.conf`.

Later, a software like a CLI will want to rewrite a key and change its value. In my example use case, I'm taking a configuration file for a given piece of software, injecting it into a Docker container, and then rewriting a modified version of it into a private dir before calling a containerized version of my program.

If a human operator has to look at the modified version, I'd highly desire for them to be able to see the comments above the key they just set.

Here's a greatly simplified example:

```toml
# ===================================================================
# My rpm module
# ===================================================================

# Enable debug level logging
#debug = true

# ===================================================================
# Attached object storage
# ===================================================================
[objectstore]

#url = "https://bucketname.sfo2.digitaloceanspaces.com"
#accesskey = ""
#secret = ""


# ===================================================================
# Attached redis
# ===================================================================
[redis]

#url = "redis://password@localhost:6379"
...
```

Issue #29, opened 2014-12-06
Support multi-line literal strings


pelletier:
(from TOML doc)

**Multi-line literal
strings** are surrounded by three single quotes on each side and allow newlines.
Like literal strings, there is no escaping whatsoever. If the first character
after the opening delimiter is a newline (`0x0A`), then it is trimmed. All other
content between the delimiters is interpreted as-is without modification.

```toml
regex2 = '''I [dw]on't need \d{2} apples'''
lines  = '''
The first newline is
trimmed in raw strings.
   All other whitespace
   is preserved.
'''
```

For binary data it is recommended that you use Base64 or another suitable ASCII
or UTF-8 encoding. The handling of that encoding will be application specific.

See #27 and #28.

pelletier:
Fixed by 42e7853

PR #34, opened 2015-02-26
Changes to support #29 - Support multi-line literal strings

eanderton:
* Added error output to test_program.go
* Added multi-line literal string support to lexer
* Added multi-line string supprt to lexer
* Added unit-test for new string support
* Modified test.sh to take an optional parameter to run an individual BurntSushi test suite.
* Fixed formatting

I forgot to log that the test.sh script now pulls from master from the BurntSushi test suite in order to validate multi-line strings.  So there are a bevy of tests for TOML 0.3.0 and beyond that don't pass... yet.

### DIFF
--- a/cmd/test_program.go
+++ b/cmd/test_program.go
@@ -13,10 +13,12 @@ import (
 func main() {
 	bytes, err := ioutil.ReadAll(os.Stdin)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pelletier:
As always, awesome work! :+1:

PR #37, opened 2015-07-15
Update keys parsing

pelletier:
@eanderton with that PR, we should handle keys like TOML 0.4: `"a b"`, `a."b c"`, etc.

The passing on my machine but failing on Travis:

```
--- FAIL: TestDateOffset-2 (0.00 seconds)
	parser_test.go:26: was expecting 1979-05-27 00:32:00 -0700 PDT at a but got 1979-05-27 00:32:00 -0700 -0700
=== RUN TestDateNano-2
--- FAIL: TestDateNano-2 (0.00 seconds)
	parser_test.go:26: was expecting 1979-05-27 00:32:00.999999999 -0700 PDT at a but got 1979-05-27 00:32:00.999999999 -0700 -0700
```

But I think that's unrelated to the changes.

### DIFF
--- a/keysparsing.go
+++ b/keysparsing.go
@@ -5,6 +5,7 @@ package toml
 import (
 	"bytes"
 	"fmt"
+	"unicode"
 )
 
 func parseKey(key string) ([]string, error) {
@@ -51,5 +52,5 @@ func parseKey(key string) ([]string, error) {
 }
 
 func isValidBareChar(r rune) bool {
-	return isAlphanumeric(r) || r == '-'
+	return isAlphanumeric(r) || r == '-' || unicode.IsNumber(r)
 }
--- a/keysparsing.go
+++ b/keysparsing.go
@@ -1,0 +1,55 @@
+// Parsing keys handling both bare and quoted keys.
+
+package toml
+
+import (
+	"bytes"
+	"fmt"
+)
+
+func parseKey(key string) ([]string, error) {
+	groups := []string{}
+	var buffer bytes.Buffer
+	inQuotes := false
+	escapeNext := false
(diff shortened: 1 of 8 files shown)


eanderton:
Looks like it's having a hard time with timestamps and timezones; this suggests a difference in platform.  @pelletier , are you developing on something other than the Linux distro that Travis uses?  I might have seen the same thing since I'm on OSX at home.

pelletier:
@eanderton I am developing on OS X too. In my opinion it looks like a Go issue, not a go-toml one. So here is my stab at avoiding that: https://github.com/pelletier/go-toml/commit/36e11971902bed4e1a3236020ca93861a3760460 What do you think?

pelletier:
After a bit of investigation, it does not seem like a platform issue, but rather a location issue. `time.Parse` takes into account the system's location, and that's why the output differs. I think we did not notice that on our machines because I bet we are both based on the US West coast. Re-reading http://golang.org/pkg/time/#Parse, I think that parsing all dates and times in UTC is the best move. The user can still change the location after the parsing using http://golang.org/pkg/time/#Time.In.

I'll merge this PR after your review.

eanderton:
It's funny you should say that - I'm actually on the East Coast.  That said, sticking to UTC hasn't steered me wrong before.  It sure beats stumbling over timezone issues.

This PR looks good.  Ship it!

pelletier:
Ah, then I must have misunderstood something in the way `time.Parse` works. :confused:  I like the idea of not dealing with timezones and deferring that job to the user.

pelletier:
@eanderton by the way, I think we are really close to support TOML v0.4.0 with that patch.

PR #38, opened 2015-07-17
Reject full 00 - 1F unicode range

pelletier:
That should give us full support for TOML v0.4 multiline @eanderton.


eanderton:
Ship it!

PR #40, opened 2015-07-17
Accept spaces in keys

pelletier:
Whitespace around dot-separated parts is ignored, however, best practice is to not use any extraneous whitespace.

```toml
[a.b.c]          # this is best practice
[ d.e.f ]        # same as [d.e.f]
[ g .  h  . i ]  # same as [g.h.i]
[ j . "ʞ" . l ]  # same as [j."ʞ".l]
```

I may have lost the track during the implementation of this one, there may be a simpler solution.

@eanderton

### DIFF
--- a/keysparsing.go
+++ b/keysparsing.go
@@ -13,7 +13,16 @@ func parseKey(key string) ([]string, error) {
 	var buffer bytes.Buffer
 	inQuotes := false
 	escapeNext := false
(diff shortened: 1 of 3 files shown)


eanderton:
If key parsing becomes any more sophisticated, it may make sense to converge all of the key parsers we have on to the same implementation, if it makes sense to do so.

Ship it!

PR #43, opened 2015-07-18
Add support for index lookups using GetPath with Groups.

gdamore:
This PR addresses two issues.

First, when using a [[group]], the library panics if there is already a different element with the same name, such as [group].  This is a configuration file error, and should not panic.  The fix for that is in parser.go.

The second problem, is really more of an enhancement.  This lets me use Get() (or GetPath()) with such groups by including an index.  That is, I can do Get("group.0.name") to get the item at index 0 (and you can use different indices.)  The old limitation of only choosing the last item in the group is removed, although at some level this change breaks that old limited behavior.

I've included a test case for this new functionality as well.  I've added my name to the copyright list, so and I contribute my changes under the same terms as the original.

I hope others find this helpful.  Let me know if there is anything else needed here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ You can run both of them using `./test.sh`.
 
 ## License
 
-Copyright (c) 2013 - 2015 Thomas Pelletier, Eric Anderton
+Copyright (c) 2013 - 2015 Thomas Pelletier, Eric Anderton, Garrett D'Amore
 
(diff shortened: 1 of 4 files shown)


pelletier:
The build is not passing because some tests related to Groups parsing are failing. Do you mind taking a look please?

pelletier:
Closing this as there as been no update in 9 months, and the code base now implements those bug fixes.

PR #56, opened 2016-04-18
Clean up lint

moorereason:
The only real change in this commit is that MaxInt is made private.
Everything else should be gofmt'ing, docs and cleanup of lint.

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -299,7 +299,7 @@ func (l *tomlLexer) lexKey() tomlLexStateFn {
 
 func (l *tomlLexer) lexComment() tomlLexStateFn {
 	for next := l.peek(); next != '\n' && next != eof; next = l.peek() {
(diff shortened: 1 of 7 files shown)


pelletier:
@moorereason Thank you for your PR! Just one issue though: [this line](https://github.com/pelletier/go-toml/pull/56/files#diff-2ea985d6df96843c96ae0340ce5faa90R356) does not compile, as `q` is scoped to the `if/else` construct ([minimal test case](https://play.golang.org/p/DJ19ApwTBP)).

Once that is issue is fixed, I'll be happy to merge this!

moorereason:
My bad.  Got in a hurry and didn't run tests!  Fixed and rebased.

pelletier:
Awesome, thanks! Merging that – disregard coveralls, which shouldn't use github statuses.

PR #59, opened 2016-04-19
TomlTree.ToMap

pelletier:
Provide a `TomlTree.ToMap()` function that generates an actual map from a `TomlTree` (obtained after a `Load*` operation.

### DIFF
--- a/toml.go
+++ b/toml.go
@@ -6,9 +6,7 @@ import (
 	"io"
 	"os"
 	"runtime"
-	"strconv"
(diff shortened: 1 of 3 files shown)


robmuh:
Plugged in the `ToMap()` PR branch and so far working well for what I need. I'm writing a `tomap_test.go` with [this example file](https://gist.github.com/robmuh/7966da29024c075349a963840e2298b2). Will let you know how it goes.

PR #67, opened 2016-04-22
Fix inline tables parsing

pelletier:
Inline tables were wrapped inside a TomlValue, although they should
just be part of the tree.


robmuh:
Sorry not seeing a change here. Pretty sure I changed the branches correctly:

screen shot 2016-04-22 at 10 02 18 am
screen shot 2016-04-22 at 10 03 37 am

robmuh:
Yep. Looks like that got it to me.
screen shot 2016-04-22 at 10 10 40 am

PR #69, opened 2016-04-22
simple tomll linter util

robmuh:
Just a nice convenience.

PR #74, opened 2016-06-06
Add a toml linter

sambooo:
Supports stdin/stdout and also writing to files. Errors will be printed to stderr so they don't get piped to files unintentionally.

Based on #69

### DIFF
--- a/cmd/tomll/main.go
+++ b/cmd/tomll/main.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"flag"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/pelletier/go-toml"
+)
+
+func main() {
+	flag.Parse()
+	// read from stdin and print to stdout
+	if flag.NArg() == 0 {
+		s, err := lintReader(os.Stdin)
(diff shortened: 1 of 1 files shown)


cmd/tomll/main.go
```diff
@@ -0,0 +1,49 @@
...
+	flag.Parse()
+	// read from stdin and print to stdout
+	if flag.NArg() == 0 {
+		s, err := lintReader(os.Stdin)
+		if err != nil {
+			io.WriteString(os.Stderr, err.Error())
+			os.Exit(-1)
+		}
+		io.WriteString(os.Stdout, s)
+		os.Exit(0)
```
pelletier: I'd rather see a `if/else` block here rather than `os.Exit(0)`.

pelletier:
That's great! Do you mind adding a `--help` flag that describe the purpose and usage of the program? I think it's important to mention that files will be modified in place.

sambooo:
Usage() is called here by flag when -h or some invalid flags are passed

pelletier:
Perfect. Thanks for your contribution!

PR #78, opened 2016-07-20
Pointer cleanup

### LINKED ISSUE #71: why are there pointer receiver methods on toml.Position?
Every function returns a value of `toml.Pointer`, not a pointer to `toml.Pointer` which makes calling them very awkward.

As in you can't do this:
```go
log.Fatalf("password must be a string: %s", t.GetPosition("password"))
```

You must instead do:
```go
p := t.GetPosition("password")
log.Fatalf("password must be a string: %s", &p)
```

This is because the `Position.String()` method is defined with a pointer receiver, and thus not part of the method set of the value struct.

sambooo:
Fixes #71 along with some other pointers

### DIFF
--- a/position.go
+++ b/position.go
@@ -18,12 +18,12 @@ type Position struct {
 
 // String representation of the position.
 // Displays 1-indexed line and column numbers.
-func (p *Position) String() string {
+func (p Position) String() string {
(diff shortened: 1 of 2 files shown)


pelletier:
Awesome! Thank you for your contribution.

Issue #83, opened 2016-08-13
Array of Tables not loaded propery


vladimirvivien:
I ran across this issue where array of tables were not converted properly.
```
[[menu.main]]
a = menu1
b = menu 2
[[menu.main]]
c = menu 3
d = menu 4
```
Instead of getting a map with key main setup as follows:
`menu:map[main:[menu1, menu2, menu3, menu4]]`
I get a key main where its elements is collapsed to only the last two items:
`menu:map[main:[menu3, menu4]]`

The code is not returning a properly constructed slice when the array of tables items have the same key names.

vladimirvivien:
@pelletier Hi, please see PR #84 .  It would be great if you could merge this soon as you can.  Thanks.

pelletier:
Hi @vladimirvivien, thank you for your feedback!

I'm confused about the expectations here. According to [toml#array-of-tables](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.4.0.md#array-of-tables), the following toml doc:

```toml
[[menu.main]]
a = "menu 1"
b = "menu 2"
[[menu.main]]
c = "menu 3"
d = "menu 4"
```

should produce the following tree:

```json
{
  "menu": {
    "main": [
      {"a": "menu 1", "b": "menu 2"},
      {"c": "menu 3", "d": "menu 4"}
    ]
  }
}
```

which is what is currently produced by go-toml.

Do you mind clarifying the issue and providing a failing test case? I'm sure we can get this fixed quickly :)

vladimirvivien:
Hi @pelletier 
Thanks for the quick reply. The problem I ran into is that example was producing:
```
{
  "menu": {
    "main": [
      {"c": "menu 3", "d": "menu 4"}
    ]
  }
}
```
Backstory: My team is trying to lobby `spf13` to adopt your toml api (due to better licensing for our use).  The problem came from the following sample from spf13/Hugo configuration.  It was failing where only the last menu set was being returned.  Also I didn't see a test case in parser_test.go for this exact usecase.

```
[[menu.main]]
    name = "Tutorials"
    identifier = "tutorials"
    pre = "<i class='fa fa-book'></i>"
    weight = -40
[[menu.main]]
    name = "Troubleshooting"
    identifier = "troubleshooting"
    pre = "<i class='fa fa-wrench'></i>"
    weight = -30
[[menu.main]]
    name = "Tools"
    url = "/tools/"
[...]

PR #84, opened 2016-08-13
Fix for Array of Tables

vladimirvivien:
The code in this PR addresses the bug documented in #83. It fixes the issue where the code is not returning a properly constructed slice when the array of tables items have the same key names.


pelletier:
Thanks! Merged with https://github.com/pelletier/go-toml/pull/86

PR #88, opened 2016-08-20
Add \U support and more tests to queryLexer

moorereason:
Increases coverage of querylexer.go from 71.6% to 97.5%.


pelletier:
Thank you for this patch!

PR #89, opened 2016-08-20
Update Travis build

moorereason:
Test with the latest releases.  Allow tip to fail.  Move script section.
Standardize whitespace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
 language: go
-script: "./test.sh"
 go:
-    - 1.5
-    - 1.6
-    - 1.7
-    - tip
+  - 1.5.4
+  - 1.6.3
+  - 1.7
+  - tip
+matrix:
+  allow_failures:
+    - go: tip
+  fast_finish: true
+script:
+  - ./test.sh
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,13 +1,18 @@
 language: go
-script: "./test.sh"
 go:
-    - 1.5
-    - 1.6
-    - 1.7
-    - tip
+  - 1.5.4
+  - 1.6.3
```
pelletier: Any reason why you would like to specify the go patch level? As far as I understand [gimme](https://github.com/travis-ci/gimme) will provide the latest patch level if it's not provided.

moorereason: That's not how gimme works.  You have to specify the version explicitly.  See the [latest build](https://travis-ci.org/pelletier/go-toml/jobs/154093148#L127) and notice that it builds with go1.6.  Compare it to [this PR build](https://travis-ci.org/pelletier/go-toml/jobs/153802919#L135).

pelletier: Ah, my bad. Just tried it out and you're right.

PR #95, opened 2016-08-22
Simplify code

moorereason:
Using `gofmt -s`


moorereason:
Most of these are pretty apparent, but the edits to `parser_test.go` are a little harder to recognize at a glance.  Let me know if you don't like it.

pelletier:
Thanks! I like it :) Today I learn that the compiler is actually able to infer those types.

Issue #96, opened 2016-09-04
empty quoted keys should be allowed.


nhooyr:
See toml-lang/toml#379.

```toml
[""]
"" = 1
```

Causes an error with go-toml even though it is valid toml.

pelletier:
Thanks! Good catch.

PR #97, opened 2016-09-06
Allow empty quoted keys

pelletier:
Allow the following:

```toml
[""]
"" = 1
```

Should fix https://github.com/pelletier/go-toml/issues/96


nhooyr:
How would `tree.GetPos` work with this if it thinks an empy key refers to the tree itself?

pelletier:
Quoted keys are also supported by `TomlTree.Get` and `TomlTree.GetPosition`. So in order to get that value, you'd need to do ```tree.Get(`"".""`)```

PR #98, opened 2016-09-10
Move token and lexer to separate packages

moorereason:
This commit is massive, but relatively simple.  It moves the main token and lexer into their own sub-packages.


pelletier:
I'm iffy about that change because it means making public a lot of internal functions, which would make changing without breaking backward compatibility more difficult in the future.

What's the motivation to pull those into sub-packages?

moorereason:
Ultimately, my goal is to pattern a token, lexer/scanner, ast, and parser (and printer) after Go.  I should probably have discussed this with you beforehand, but I figured you would ask such as question.  :blush:

One of the main motivators is that I want to be able to read in a TOML file, modify a field in the AST, and then output the results while maintaining comments and the general structure (as in, don't sort the keys or lose comments in output).  I would like to have a `tomlfmt` in the end.

But having said all of that, you've raised a concern about the API that I hadn't considered.  I was trying to make this move without changing too many things at once, but I think that would only create more problems.  Tunnel vision.

I'll close this PR and rethink the API of these two packages.  If you're not interested in where I'm headed, just let me down easy.  :wink:

Issue #101, opened 2016-09-19
String() fails on big integers


bep:
```
unsupported value type: 6077473420639406657
```

bep:
Great! Thanks for the

moorereason:
Team Hugo FTW!  :wink:

PR #103, opened 2016-09-19
Handle nil, map[string]string, and map[interface{}]interface{} input

### LINKED ISSUE #99: String() fails on map[string]string
```go
package main

import (
	"fmt"

	toml "github.com/pelletier/go-toml"
)

func main() {
	m1 := map[string]string{"v": "abc"}
	m2 := map[string]interface{}{"m": m1}
	tree := toml.TreeFromMap(m2)
	fmt.Println(tree.String())
}
```

Panics:

```
panic: unsupported value type: map[v:abc]
```

In Hugo we do frontmatter conversions between YAML and TOML etc., and the datatypes for maps differs slightly. We recently converted from Burntsushi to pelletier's go-toml for Hugo and related libraries. We are in the process to release a new version of Hugo and need to fix this in a way or another.

See https://github.com/spf13/hugo/issues/2439

moorereason:
Fixes #99

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -658,6 +658,28 @@ func TestToString(t *testing.T) {
 	}
 }
 
+func TestToStringMapStringString(t *testing.T) {
+	in := map[string]interface{}{"m": map[string]string{"v": "abc"}}
+	want := "\n[m]\n  v = \"abc\"\n"
+	tree := TreeFromMap(in)
+	got := tree.String()
+
+	if got != want {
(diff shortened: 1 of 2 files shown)


pelletier:
@moorereason thank you for taking care of this and #101! :)
I think those two issues appeared because the TomlTree is being created in a way it wasn't intended to. I think we should find a better solution to allow converting a golang structure to a TomlTree (and back) in a more robust fashion.

moorereason:
Agreed.

PR #117, opened 2016-12-03
Make ToString() return an error instead of panic

pelletier:


pelletier:
@bep @moorereason thoughts? Not very fond of the catch-all recover, but I wrote a first version handling the errors manually, and it made the code unnecessarily difficult to read.

PR #126, opened 2017-01-21
Fix syntax errors in package-level documentation

docbrown:


pelletier:
Good catch, thank you!

PR #123, opened 2017-01-07
Add WriteToToml to TomlTree

tgulacsi:
This is a more efficient, WriterTo-like implementation,
with key sorting for reproducibility.

PR #133, opened 2017-03-02
Rewrite TomlTree encoding

pelletier:
Notable changes:

* Encoding makes the assumption that the tree is well formed, meaning that it is only composed of those types for leaf nodes: `string`, `int64`, `uint64`, `float64`, `bool`, `time.Time`, `[] of any of this list`; and those for intermediate nodes: `*tomlValue`, `*tomlTree`, `[]*TomlTree`. This allows for a great deal of simplification in the encoding code, as well as saner logic. The caveat is that `TreeFromMap()` currently does not perform any sanitization. I'm planning on implementing that in subsequent changes, but in the meantime the tree should be properly constructed using `Set()`.
* Encoding sorts keys by alphabetical order.
* `tomltree_conversions*.go` files are now `tomltree_write*.go`.
* A new method, `tree.WriteTo(io.Writer) (int64, error)` is now available, for all the encoding needs.

@tgulacsi this is loosely based on your https://github.com/pelletier/go-toml/pull/123. Thanks for the work and support!


tgulacsi:
LGTM

PR #139, opened 2017-03-09
Rework tree from map

pelletier:
This is a complete rework of `TreeFromMap()`. It now constructs a well formed tree and returns an error if the data cannot be matched to a known TOML type.

### DIFF
--- a/toml.go
+++ b/toml.go
@@ -28,10 +28,16 @@ func newTomlTree() *TomlTree {
 }
 
 // TreeFromMap initializes a new TomlTree object using the given map.
-func TreeFromMap(m map[string]interface{}) *TomlTree {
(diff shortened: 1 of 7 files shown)


toml.go
```diff
@@ -28,10 +28,12 @@ func newTomlTree() *TomlTree {
 }
 
 // TreeFromMap initializes a new TomlTree object using the given map.
-func TreeFromMap(m map[string]interface{}) *TomlTree {
-	return &TomlTree{
-		values: m,
+func TreeFromMap(m map[string]interface{}) (*TomlTree, error) {
```
raphael: FYI this change breaks hugo and I'm sure other dependent packages :(

PR #140, opened 2017-03-14
Support lowercase unicode escape sequences

moorereason:
As seen here: https://discuss.gohugo.io/t/5789


pelletier:
Good catch, completely missed that from the spec. Thank you!

Issue #146, opened 2017-03-21
Add capability for reflection-based (un)marshalling


tro3:
The golang/dep project is hoping to use this library, and the ability to marshal/unmarshal directly into dep structures would be helpful.  WIP PR for early feedback forthcoming.

PR #149, opened 2017-03-21
Reflection-based marshaling / unmarshaling

tro3:
Going after #146,  @pelletier this initial commit is just to give you a feel for the direction I am heading with this.  Still need to add Map and Array handling, error handling (with position info for unmarshaling), and lots of tests.  Please let me know your thoughts, so I can shift course early if needed.

One question - do you want to follow the `encoding/json` pattern of ignoring non-exported fields?

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -1,0 +1,163 @@
+package toml
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// Marshal ...
+func Marshal(v interface{}) ([]byte, error) {
+	vt := reflect.TypeOf(v)
+	if vt.Kind() != reflect.Struct {
+		return []byte{}, errors.New("Only a Struct can be marshaled to TOML")
+	}
+	vv := reflect.ValueOf(v)
+	t := newTomlTree()
+	path := []string{}
+	err := reflectTreeFromStruct(vt, vv, t, path)
+	if err != nil {
+		return []byte{}, err
+	}
+	s, err := t.ToTomlString()
+	return []byte(s), err
+}
+
+func reflectTreeFromStruct(vt reflect.Type, vv reflect.Value, t *TomlTree, path []string) error {
+	for i := 0; i < vt.NumField(); i++ {
+		vtf, vvf := vt.Field(i), vv.Field(i)
+		keys := append(path, tomlName(vtf))
+		err := reflectTreeFromValue(vtf.Type, vvf, t, keys)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func reflectTreeFromValue(vt reflect.Type, vv reflect.Value, t *TomlTree, path []string) error {
+	switch vt.Kind() {
+	case reflect.Bool:
+		t.SetPath(path, vv.Bool())
+	case reflect.Int:
+		fallthrough
+	case reflect.Int8:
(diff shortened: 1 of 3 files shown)


pelletier:
Good job! That's pretty much what I had in mind. I'd like to stay as close as possible to the behavior of `encoding/json`. So yes, I would ignore non-exported fields.

tro3:
@pelletier - I've removed the WIP tag.  This is ready for review.  The only feature it does not have is the handling of `interface{}` types in the object being marshalled.  That is a big enough job (I think) to warrant another PR.

@carolynvs, @sdboyer - Would love to hear feedback on how usable this appears in the `dep` case.

tro3:
Hmm, it occurs I still need documentation...   Will add.

carolynvs:
@tro3 I think this is pretty close! 🎉 I would just need `omitempty` and could replace the manual mapping in my PR.

tro3:
I'm on it.

tro3:
@carolynvs - `omitempty` is functional.  @pelletier - ready for review.  Note to both that I have made pointers automatically `omitempty`, as TOML wants null values dropped from the file, per https://github.com/toml-lang/toml/issues/30

marshal.go
```diff
@@ -0,0 +1,477 @@
...
+			mtypef := mtype.Field(i)
+			opts := tomlOptions(mtypef)
+			if opts.include {
+				key := opts.name
+				exists := tval.Has(key)
+				if exists {
+					val := tval.Get(key)
+					mvalf, err := valueFromToml(mtypef.Type, val)
+					if err != nil {
+						if err.Error()[0] == '(' {
```
pelletier: Seems weird how to do some logic based off the first character of the error message. Do you mind elaborating on how this works?

marshal.go
```diff
@@ -0,0 +1,477 @@
...
+(i.e. no `interface{}`).
+
+Note that pointers are automatically assigned the "omitempty" option, as TOML
+explicity does not handle null values (saying instead the label should be
+dropped).
+*/
+func Marshal(v interface{}) ([]byte, error) {
+	mtype := reflect.TypeOf(v)
+	if mtype.Kind() != reflect.Struct {
+		return []byte{}, errors.New("Only a Struct can be marshaled to TOML")
```
pelletier: `struct`, lower `s` :)

tro3: will fix

marshal.go
tro3: This is just to prevent every level of hierarchy adding its own location info.  I'll take a look to see if there is a cleaner error handling method.

tro3:
Will work on the above and add test coverage

marshal.go
pelletier: Got it. If you don't find a cleaner way to do it, I think just adding a comment in places you do that is good enough for now. There is the broader case of how we report errors in toml that needs to be addressed too.

tro3: Agreed.  I've already moved to a separate function `formatError` with a comment describing the "why".  Easier to rework later, if needed.  Just fighting test coverage now - forcing `omitempty` onto pointers has resulted in some unreachable(/untestable) code.

tro3:
@pelletier - Incorporated the feedback and did indeed find some holes due to lack of test coverage.   Let me know what more is needed.

PR #151, opened 2017-04-04
Use MarshalTOML() if struct has defined a custom marshaler

carolynvs:
I based `tom.Marshaler` on the implementation for `json.Marshaler`.


pelletier:
This is very nice! Thank you for the patch.

PR #155, opened 2017-04-16
Initialize keys array to final length

kevinburke:
Previously we'd create an empty array and need to continuously resize
it as we appended more entries. This way we immediately create the
correct size array, and then add entries to it.


pelletier:
Good catch, thank you!

PR #159, opened 2017-05-08
Rename TomlTree to Tree

pelletier:
@moorereason @carolynvs @bep @eanderton  – and probably many others I have forgotten.

The patch renames `toml.TomlTree` to `toml.Tree`. This is the last backward-compatibility breaking change that I want to get in before tagging v1.0.0. Unless anyone has a strong case against it, I'm planning to go forward with that. :)


carolynvs:
Yay! 🎉

PR #162, opened 2017-05-16
Fix Tree.ToMap godoc comment

### LINKED ISSUE #160: doco: float64 missing in list of types
so far loving go-toml, but found a nit...

in https://github.com/pelletier/go-toml/blob/master/tomltree_write.go#L190

i don't see `float64` in this list, and note that the [online godoc](https://godoc.org/github.com/pelletier/go-toml#Tree.ToMap)  destroys your list and turns into a mess.

"* unit64 *int64 *bool ...."

```
// ToMap recursively generates a representation of the tree using Go built-in structures.
// The following types are used:
// * uint64
// * int64
// * bool
// * string
// * time.Time
// * map[string]interface{} (where interface{} is any of this list)
// * []interface{} (where interface{} is any of this list)
```

moorereason:
Fixes #160

### DIFF
--- a/tomltree_write.go
+++ b/tomltree_write.go
@@ -189,13 +189,15 @@ func (t *Tree) String() string {
 
 // ToMap recursively generates a representation of the tree using Go built-in structures.
 // The following types are used:
-// * uint64
-// * int64
-// * bool
-// * string
-// * time.Time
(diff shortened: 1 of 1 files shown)


pelletier:
That was fast, thank you!

PR #164, opened 2017-05-25
switch kindToTypeMapping from map to more performant array

jmank88:
This change swaps out the `kindToTypeMapping` map for an array, since the `reflect.Kind` keys are just `uint` `iota`s. Obviously it's already a very fast operation, but a quick benchmark showed a significant speedup factor from `14.5 ns/op` to `1.77 ns/op`.

### DIFF
--- a/tomltree_create.go
+++ b/tomltree_create.go
@@ -6,10 +6,7 @@ import (
 	"time"
 )
 
-// supported values:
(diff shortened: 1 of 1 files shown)


pelletier:
Very nice! Thank you for the patch.

PR #166, opened 2017-05-30
Use bytes.Buffer for tomlLexer.buffer

### LINKED ISSUE #165: The lexer allocates too much
The TOML lexer lists fairly high on my profiling session in Hugo:

```
         .          .     52:func (l *tomlLexer) next() rune {
         .      10923     53:	r := l.read()
         .          .     54:
         .          .     55:	if r != eof {
   3825735    3825735     56:		l.buffer = append(l.buffer, r)
         .          .     57:	}
         .          .     58:	return r
         .          .     59:}
         .          .     60:
         .          .     61:func (l *tomlLexer) ignore() {
```

The above is allocated objects. I would:

1) Use `bytes.Buffer` (it has a `WriteRune` method).
2) Create a `sync.Pool` for those buffers.

bogem:
Fix #165

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -6,6 +6,7 @@
 package toml
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -24,7 +25,7 @@ type tomlLexStateFn func() tomlLexStateFn
 // Define lexer
 type tomlLexer struct {
 	input         *buffruneio.Reader // Textual source
-	buffer        []rune             // Runes composing the current token
+	buffer        bytes.Buffer       // Runes composing the current token
 	tokens        chan token
 	depth         int
 	line          int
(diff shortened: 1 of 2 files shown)


kevinburke:
Can you post before/after benchmarks? Or better yet amend the commit and put them in there?

bogem:
name     old time/op    new time/op    delta
Lexer-4     343µs ± 1%     331µs ± 1%  -3.56%  (p=0.000 n=20+19)

name     old alloc/op   new alloc/op   delta
Lexer-4    55.8kB ± 0%    50.8kB ± 0%  -8.86%  (p=0.000 n=20+20)

name     old allocs/op  new allocs/op  delta
Lexer-4     2.01k ± 0%     1.84k ± 0%  -8.46%  (p=0.000 n=20+20)

pelletier:
Thank you very much for tackling that!

Issue #167, opened 2017-05-30
Performance issues


bep:
Let us create a more general issue to track this.

See https://github.com/spf13/hugo/issues/3541

That issue shows a significant penalty in just switching from YAML to TOML as front matter in Hugo, both in processing time and memory allocations. I'm not saying it needs to be _as fast_, TOML is a more convenient format, but I think this needs to be looked into.

I think this library needs some simple and isolated benchmarks that compare it to, say, YAML and JSON.

pelletier:
I agree. I haven't paid any attention to performance up until now. There are probably a bunch of low hanging improvements. Do you have some real world example files that we could use for initial benchmarks?

bep:
>Do you have some real world example files that we could use for initial benchmarks?

No.

kevinburke:
For this function:

```go
func encodeTomlString(value string) string {
	result := ""
	for _, rr := range value {
		switch rr {
		case '\b':
			result += "\\b"
		case '\t':
			result += "\\t"
		case '\n':
			result += "\\n"
		case '\f':
			result += "\\f"
		case '\r':
			result += "\\r"
		case '"':
			result += "\\\""
		case '\\':
			result += "\\\\"
		default:
			intRr := uint16(rr)
			if intRr < 0x001F {
				result += fmt.Sprintf("\\u%0.4X", intRr)
			} else {
				result += string(rr)
			}
		}
	}
	return result
}
```

You can probably get a speedup by creating a `map[utf8.RuneSelf]bool` table and checking whether a character exists in the map. If it doesn't you can `continue` early. That would beat the switch statement which forces 7 comparisons before executing the `default` code block.

bep:
What I would do, before starting to make random changes that look like hot spots, is this:

1) Create a couple of benchmarks with a fairly simple, but representable, document -- converted to TOML, YAML and JSON (there are online converters)
2) Run the benchmark and output mem and cpu profiles. Looking at it with `go tool pprof --alloc_objects` is a good start -- memory allocations light up like a Christmas tree, and there is usually a strong correlation from that to speed.
3) Fix the hot spots, repeat until you are within n % of YAML/JSON, or whatever.

pelletier:
Putting https://github.com/spf13/hugo/pull/3545 here for future reference.

moorereason:
Copying a comment @pelletier [made elsewhere](https://github.com/pelletier/go-toml/pull/438#issuecomment-691509873) since it's relevent to this issue:

[...]

PR #170, opened 2017-05-31
Optimize some string handling

moorereason:
I've added an additional commit to this PR that further optimizes this area of code:

```none
benchmark                       old ns/op     new ns/op     delta
BenchmarkTreeToTomlString-2     49028         18817         -61.62%

benchmark                       old allocs     new allocs     delta
BenchmarkTreeToTomlString-2     485            78             -83.92%

benchmark                       old bytes     new bytes     delta
BenchmarkTreeToTomlString-2     9749          4961          -49.11%
```

### DIFF
--- a/tomltree_write.go
+++ b/tomltree_write.go
@@ -13,33 +13,34 @@ import (
 
 // encodes a string to a TOML-compliant string value
 func encodeTomlString(value string) string {
-	result := ""
+	var b bytes.Buffer
+
 	for _, rr := range value {
 		switch rr {
 		case '\b':
-			result += "\\b"
+			b.WriteString(`\b`)
 		case '\t':
-			result += "\\t"
+			b.WriteString(`\t`)
 		case '\n':
-			result += "\\n"
+			b.WriteString(`\n`)
 		case '\f':
-			result += "\\f"
+			b.WriteString(`\f`)
 		case '\r':
-			result += "\\r"
+			b.WriteString(`\r`)
 		case '"':
-			result += "\\\""
+			b.WriteString(`\"`)
 		case '\\':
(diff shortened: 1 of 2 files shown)


tomltree_write.go
```diff
@@ -111,7 +111,7 @@ func (t *Tree) writeTo(w io.Writer, indent, keyspace string, bytesCount int64) (
 			return bytesCount, err
 		}
 
-		kvRepr := fmt.Sprintf("%s%s = %s\n", indent, k, repr)
+		kvRepr := indent + k + " = " + repr + "\n"
```
kevinburke: You could probably speed this up even further by creating a bytes.Buffer outside of the loop range, writing to it inside the for loop, then calling `w.Write(buf.Bytes())`, which would also skip the string conversion below this line.

moorereason: Tried your suggestion, but it only added allocs. Perhaps I'm doing it wrong.

kevinburke:
Here's what I had in mind. It makes the code a lot simpler, since we know that `b.WriteString` and company never return an error, we don't need to check them constantly.

https://github.com/pelletier/go-toml/commit/b49b944a707ec81a11ce2f71249f8cf2bc32b5e4

pelletier:
@moorereason nice!
@kevinburke I like your proposition (makes the code much cleaner), but the +14% allocs increase isn't great. What happens with a larger preallocated buffer?

I think I'm going to merge this as a first step. Then a second step we should figure a way to make Kevin's solution use some less allocs, as it makes the code much more readable.

pelletier:
Ref https://github.com/pelletier/go-toml/issues/167

PR #175, opened 2017-06-02
Initial benchmark against other libraries

pelletier:
Compare go-toml to json and yaml. First results:

```
BenchmarkUnmarshalToml-8             1000  1320117 ns/op  450932 B/op  15072 allocs/op
BenchmarkUnmarshalBurntSushiToml-8   3000   402897 ns/op   82900 B/op   1761 allocs/op
BenchmarkUnmarshalJson-8            20000    66092 ns/op    3536 B/op    101 allocs/op
BenchmarkUnmarshalYaml-8            10000   189600 ns/op   44872 B/op   1058 allocs/op
```

Plenty of room for improvement.

Related to https://github.com/pelletier/go-toml/issues/167

@bep @kevinburke @moorereason

### DIFF
--- a/benchmark.json
+++ b/benchmark.json
@@ -1,0 +1,164 @@
+{
+    "array": {
+        "key1": [
+            1,
+            2,
+            3
+        ],
+        "key2": [
+            "red",
+            "yellow",
+            "green"
+        ],
+        "key3": [
+            [
+                1,
+                2
+            ],
+            [
+                3,
+                4,
+                5
+            ]
+        ],
+        "key4": [
+            [
+                1,
+                2
+            ],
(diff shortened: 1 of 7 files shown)


bep:
I would also add BurntSushi's TOML to the mix, it seem to be pretty fast.

kevinburke:
Thats' also not really a fair comparison, the toml library preserves much more data about the underlying file (so it can be accurately reproduced)

pelletier:
@bep Good idea.
@kevinburke Although that's the case in the current implementation, I don't think Unmarshal should do that, given that this information would be lost anyway.

PR #178, opened 2017-08-11
Fix examples

178inaba:
Fix to define package name.
Add Output test.
Add ExampleMarshal.


pelletier:
Good catch! Thank very much for the fix.

PR #190, opened 2017-10-06
Marshal map[string]interface{} with quoted keys

robertgzr:
The TOML spec supports using UTF-8 strings as keys.
https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.4.0.md#table

This PR changes the Marshal function to output quoted strings if it
encounters a map[string]interface{}

Usage example:
Currently the following example can not be unmarshaled -> manipulated -> marshaled
```go
type Config struct {
  Pathmap map[string]string
}
```
```toml
[pathmap]
  "/path/to/somewhere" = "/path/to/somewhere/else"
```

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -167,7 +167,11 @@ func valueToTree(mtype reflect.Type, mval reflect.Value) (*Tree, error) {
 			if err != nil {
 				return nil, err
 			}
-			tval.Set(key.String(), "", false, val)
+			keyStr, err := tomlValueStringRepresentation(key.String())
+			if err != nil {
+				return nil, err
+			}
+			tval.SetPath([]string{keyStr}, "", false, val)
 		}
 	}
 	return tval, nil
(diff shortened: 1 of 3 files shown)


moorereason:
This looks too naive.  As long as the key contains characters allowed by bare keys, there's no reason to quote it.

> Bare keys may only contain letters, numbers, underscores, and dashes (A-Za-z0-9_-).

robertgzr:
@moorereason I did it like this because TOML bare keys match up with the rules for identifiers in Go.

Another option would be to have some kind of option system like the stdlib has for [`encoding/json`](https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L328)

robertgzr:
I updated the PR to avoid issues with the key parser interpreting `/path/to/file.toml` as a complex key path: 
```toml
[map]
  [path/to/file]
    toml = "..."
```

robertgzr:
@pelletier any input on this?

pelletier:
I agree with @moorereason: bare keys should not be quoted. I think the best would be to mimic `encoding/json` by providing `Encoder` and `Decoder` types that would contain options, and be used by Marshal/Unmarshal using defaults.

robertgzr:
I started working on implementing it like this. Should I close this and reopen a new PR once I have something?

pelletier:
@robertgzr That'd be perfect yes, thank you!

robertgzr:
Replaced by #192

PR #192, opened 2017-10-19
Add Encoder/Decoder types

robertgzr:
Usage is similar to the stdlibs JSON encoder/decoder but I tried to
leave the general structure of the code the same.

Main motivation was to support encoding/decoding options to allow
encoding string-type map keys as quoted TOML keys.
This was implemented on the Encoder with QuoteMapKeys(bool).

> The TOML spec supports using UTF-8 strings as keys.
> https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.4.0.md#table

Replaces #190

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


robertgzr:
@pelletier @moorereason feedback?

robertgzr:
Moved the new types into `marshal.go` as suggested. Makes sense as the types are now defined where they are used...
Additionally I embedded the options in the encoder/decoder types which makes accessing them nicer.

I also made sure the package global `Unmarshal` uses the defaults.

robertgzr:
I will squash once you approve 🙂

pelletier:
@robertgzr Seems like you have some conflicts :) Don't worry about squashing, I always squash and merge.

robertgzr:
@pelletier I just wanted to make it easier to look at individual commits but it seems Github doesn't like the first one :P 
I will just force rebase if you don't mind

pelletier:
Go for it!

pelletier:
Thank you very much for that contribution @robertgzr!

Issue #61, opened 2016-04-21
Single quoted keys support


pelletier:
@robmuh provided [this example file](https://gist.github.com/robmuh/7966da29024c075349a963840e2298b2). It produces errors during parsing. They need to be fixed.

pelletier:
@robmuh your example file contains single-quoted keys: https://gist.github.com/robmuh/7966da29024c075349a963840e2298b2#file-example-toml-L19-L20
That's the thing that makes go-toml return an error.

However, looking at the TOML 0.4.0 spec, I don't see anything mentioning this is valid syntax. The only thing I can find is

> Keys may be either bare or quoted. Bare keys may only contain letters, numbers, underscores, and dashes (A-Za-z0-9_-). Quoted keys follow the exact same rules as basic strings and allow you to use a much broader set of key names.

Which allows for `bare_keys` and `"double quoted keys"`, but not `'single quoted keys'.

robmuh:
Unwittingly I copied blindly the examples from the main TOML page (not knowing about toml-test, which is too out of date to be useful at this point unfortunately)

<img width="920" alt="screen shot 2016-04-21 at 8 46 38 am" src="https://cloud.githubusercontent.com/assets/3749067/14709259/bc2553ca-079d-11e6-8055-82eb53b4d3d5.png">

Those look like single quoted keys to me, (even though I will never use them personally).

I believe testing the spec remains one of the biggest challenges to TOML going forward (and I would personally like to see it submitted as an IETF RFC by the end of the year if we can manage).

pelletier:
The examples you are showing come from the latest commit on master, but not from the v0.4.0 specification.Although I'm all for trying to anticipate changes in the spec, this is not yet something that needs to be supported for the v0.4.0 compatibility.

robmuh:
After downloading the v0.4.0 release from 2015 that Tom himself made I see what you mean:
![screen shot 2016-04-21 at 1 19 41 pm](https://cloud.githubusercontent.com/assets/3749067/14718299/df096838-07c3-11e6-99ee-77cfb250c6db.png)

Please understand my frustration since—like the rest of the world—I am going off everything that is linked to the [TOML](http://github.com/toml-lang/toml) page listing the new stuff, which we just tripped over, very clearly at v0.4.0. 

[...]

PR #193, opened 2017-10-19
Support single quoted keys (#61)

kzys:
This change supports single quoted keys _as like_ quoted keys. However both don't support escape sequences correctly.

According to the spec;

> Quoted keys follow the exact same rules as either basic strings or literal strings and allow you to use a much broader set of key names.

Should we fix both of them in this PR?

### DIFF
--- a/keysparsing.go
+++ b/keysparsing.go
@@ -6,15 +6,37 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"strconv"
 	"unicode"
+)
+
+var escapeSequenceMap = map[rune]rune{
+	'b':  '\b',
+	't':  '\t',
+	'n':  '\n',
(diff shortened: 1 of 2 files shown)


moorereason:
If this PR doesn't implement single-quoted keys correctly, then yes, please fix that in this PR.

kzys:
I've updated the PR to support escape sequence correctly. The logic is pretty similar to lexString(), which should be merged into one function though.

moorereason:
@kzys, use gofmt.

kzys:
Thanks. Updated the pull request.

pelletier:
Merged https://github.com/pelletier/go-toml/commit/a410399d2c85888843bf75f4691886acdecd9d69. Thank you for implementing that!

Issue #181, opened 2017-08-25
Setup AFL fuzzer


pelletier:
The project has been fuzzed a couple times, but [go-fuzz](https://github.com/dvyukov/go-fuzz) should be easily runnable from a script in the repo.

PR #194, opened 2017-10-20
Add fuzz.sh to do fuzzing with go-fuzz (#181)

kzys:
I can't find interesting bugs so far. Probably due to the corpus?

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -1,0 +1,31 @@
+// +build gofuzz
+
+package toml
+
+func Fuzz(data []byte) int {
+	tree, err := LoadBytes(data)
+	if err != nil {
+		if tree != nil {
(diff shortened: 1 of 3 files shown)


fuzz.go
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	str, err := tree.ToTomlString()
+	if err != nil {
+		if str != "" {
+			panic(`str must be "" if there is an error`)
+		}
+		panic(err)
+	}
+	return 1
```
pelletier: maybe add one extra step to parse back `str`?

kzys: Makes sense. I've updated the PR.

pelletier:
Thank you for setting this up!

Issue #186, opened 2017-09-12
Unmarshal panic when structure field type is custom


Aukstkalnis:
I created toml file:

[[Device]]
Imei = 3500000000000
Job = 0

[[Device]]
Imei = 3500000000000
Job = 0

Code:
type Job uint8
type Device struct {
    Imei uint64
    AssignedJob Job
}
...
type DeviceList []Device
var dev DeviceList
toml.Unmarchal(file, &dev)

An I get panic:
panic: reflect.Set: value of type uint8 is not assignable to type device.Job [recovered]
        panic: reflect.Set: value of type uint8 is not assignable to type device.Job

goroutine 18 [running]:
testing.tRunner.func1(0xc0420ac0f0)
        C:/go/src/testing/testing.go:711 +0x2d9
panic(0x559e20, 0xc0420e95b0)
        C:/go/src/runtime/panic.go:491 +0x291
reflect.Value.assignTo(0x559fa0, 0x648ca0, 0x88, 0x590d16, 0xb, 0x558a20, 0x0, 0x559720, 0x679fa0, 0x0)
        C:/go/src/reflect/value.go:2186 +0x3ad
reflect.Value.Set(0x558a20, 0xc042052af8, 0x188, 0x559fa0, 0x648ca0, 0x88)
        C:/go/src/reflect/value.go:1351 +0xab
github.com/pelletier/go-toml.valueFromTree(0x649760, 0x57a540, 0xc042051840, 0xa, 0x556b20, 0xc0420f8780, 0x97, 0x649760)
        D:/Projects/Go_Projects/src/github.com/pelletier/go-toml/marshal.go:283 +0x45f
github.com/pelletier/go-toml.valueFromTreeSlice(0x649760, 0x556b20, 0xc04209e780, 0xa, 0x10, 0x411e63, 0xc0420b2ea0, 0x646a80, 0xc0420b2f78, 0xc0)
        D:/Projects/Go_Projects/src/github.com/pelletier/go-toml/marshal.go:306 +0x171
github.com/pelletier/go-toml.valueFromToml(0x649760, 0x556b20, 0x556560, 0xc0420f8600, 0xc0420f8600, 0x556b20, 0x0, 0x0, 0x0)
        D:/Projects/Go_Projects/src/github.com/pelletier/go-toml/marshal.go:337 +0x1a7
[...]

PR #196, opened 2017-10-20
Unmarshal should report a type mismatch as an error (#186)

kzys:


pelletier:
Thank you for your patch!

Issue #191, opened 2017-10-15
Unmarshal example for README doesn't work


mlcdf:
Hello!

It seems that the Unsmarshal example from the README doesn't work.

```go
type Postgres struct {
    User     string
    Password string
}
type Config struct {
    Postgres Postgres
}

doc := []byte(`
[postgres]
user = "pelletier"
password = "mypassword"`)

config := Config{}
toml.Unmarshal(doc, &config)
fmt.Println("user=", config.Postgres.User)

```

The `config` object is empty.

It may be because of the case of the struct `Postgres` doesn't match what in the toml (`postgres`).

pelletier:
Good catch. You are right: this is a case issue. Will fix that.

mlcdf:
I can work on fixing and updating this example (if you want).

pelletier:
I won't be able to take a look until this weekend, but I'd be happy to merge your patch before then!

pelletier:
Fixed in https://github.com/pelletier/go-toml/pull/197

PR #197, opened 2017-10-21
Fix example code in README

mlcdf:
Related to #191.


pelletier:
Thank you very much for that fix!

mlcdf:
You're very welcome.

Thank you for making this lib.

PR #208, opened 2018-01-14
Actually show the error message from an Error token

JelteF:
Before the error output is like this:

```
(37, 13): unexpected token type in inline table: Error
```

With this change it's like this:

```
(37, 13): unexpected token type in inline table: keys cannot contain } character
```

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -350,7 +350,7 @@ Loop:
 			}
 			p.getToken()
 		default:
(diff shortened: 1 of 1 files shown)


JelteF:
The travis failure seems to be caused by some network failure. I'm quite sure it should be fine when restarted.

PR #209, opened 2018-01-14
Unmarshal into custom types and error on overflows

JelteF:
This fixes two unmarshalling issues:

1. Unmarshalling into a custom integer/float type (e.g. `time.Duration`).
2. Checks for overflows happen before unmarshalling, erroring if an overflow
    would happen.

Apart from this it also reduces code duplication a bit.


pelletier:
Thank you for the patch!

PR #210, opened 2018-01-15
Add support for special float values (inf and nan)

JelteF:
This adds support for parsing and writing signed and unsigned versions of NaN and Inf


pelletier:
Neat. I guess I should add a list of supported TOML edge (> 0.4) features now. Thank you very much for your contribution!

PR #217, opened 2018-02-22
Make multi-line arrays always use trailing commas

ctd1500:
This makes the Encoder option `ArraysWithOneElementPerLine` output arrays with commas after every element.

```
A = [1,2,3]
```

Now becomes:

```
A = [
  1,
  2,
  3,
]
```

Originally proposed here: golang/dep#1595

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -187,7 +187,7 @@ func (e *Encoder) QuoteMapKeys(v bool) *Encoder {
 //   A = [
 //     1,
 //     2,
-//     3
+//     3,
 //   ]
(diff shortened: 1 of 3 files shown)


pelletier:
Looks good to me. Thank you for the patch!

sdboyer-stripe:
would love to see a release, so that i can roll this in with the next version of dep :)

pelletier:
@sdboyer-stripe it's been a while indeed! there you go: https://github.com/pelletier/go-toml/releases/tag/v1.2.0

PR #221, opened 2018-03-06
Added new tag 'multiline'.

alankm:
The new multiline tag works just like the existing 'commented' tag (i.e. `multiline:"true"`), and tells go-toml to marshal the value as a multi-line string. The tag currently has no impact on any non-string fields.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -11,10 +11,13 @@ import (
 	"time"
 )
 
+const tagKeyMultiline = "multiline" // alankm NOTE: key string for the new multiline struct tag
+
 type tomlOpts struct {
 	name      string
 	comment   string
 	commented bool
+	multiline bool // alankm NOTE: a new field option for multi-line strings
 	include   bool
 	omitempty bool
 }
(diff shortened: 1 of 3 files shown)


alankm:
During development for some of our projects we found that we wanted to enable users to provide substantial amounts of multi-line data in a TOML file. TOML supports this natively via triple-quotes ("""My multiline data."""), but does not require a parser to preserve this formatting. This behaviour resulted in a poor user-experience whenever we needed to modify data within a TOML file programmatically since large amounts of easily readable information would be condensed into an unreadably large single-line string with escape characters (eg: '\n') instead of real formatting.

I have addressed this issue by forking your library and adding in a new struct field tag "multiline", that carries a boolean much like your "commented" tag. This tag only has any effect on string primitives, so you may wish to review and alter the implementation to make it better fit within the library as a whole. 

I believe that none of the changes I've made break backwards compatibility, and all unit tests are succeeding. Every location in the code where I have added or modified lines is marked with a note that can be found by searching for "// alankm NOTE: ".

pelletier:
Very sorry I dropped the ball on that! Merged in https://github.com/pelletier/go-toml/commit/c01d1270ff3e442a8a57cddc1c92dc1138598194

PR #223, opened 2018-03-23
Fix Go 1.10 issues

moorereason:
This PR updates Travis CI to use the latest Go releases and fixes a couple issues introduced by changes in Go 1.10.  See commit messages for more details.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -23,9 +23,6 @@ function git_clone() {
 # Remove potential previous runs
 rm -rf src test_program_bin toml-test
 
-# Run go vet
-go vet ./...
-
 go get github.com/pelletier/go-buffruneio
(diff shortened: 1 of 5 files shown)


pelletier:
Awesome, thank you for tackling that! TIL the "1.10.x" syntax.

test.sh
```diff
@@ -23,9 +23,6 @@ function git_clone() {
 # Remove potential previous runs
 rm -rf src test_program_bin toml-test
 
-# Run go vet
-go vet ./...
```
Carpetsmoker: Note that `go test` only runs a limited subset of `vet` tests, as not all of them all considered reliable enough to be included by default (yet). So running `go test` and `go vet` is *not* identical.

PR #227, opened 2018-06-20
Fix the Windows support in file/path not found message.

hsluoyz:
The original path not found message is only for Linux: ``open /this/file/does/not/exist: no such file or directory``. But on Windows the message is: ``open /this/file/does/not/exist: The system cannot find the path specified.`` So the test will fail on Windows.

https://github.com/pelletier/go-toml/blob/31055c2ff0bb0c7f9095aec0d220aed21108121e/cmd/tomljson/main_test.go#L78-L82

### DIFF
--- a/cmd/tomljson/main_test.go
+++ b/cmd/tomljson/main_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"runtime"
 )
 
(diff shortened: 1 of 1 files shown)


pelletier:
Thank you for the fix! Reordered the the imports to make `go fmt` happy and merged here https://github.com/pelletier/go-toml/commit/603baefff989777996bf283da430d693e78eba3a.

Issue #219, opened 2018-03-02
go-toml panics if unmarshal finds what it thinks is an int while the struct has float64


willauld:
The following code panics if the number in the []byte for "percent" is 6 (not 6.0 or 6.5). The go representation must be float because it needs to be able to take on these values. However, users putting data into the file will not be so strict and Toml is much more forgiving. For this reason I think this is a bug. If the field is looking for float unmarshal should convert it as needed. 

func f3() {
	type Person struct {
		Name    string
		Age     int64
		Percent float64
	}

	document := []byte(`
		name = "John"
		age = 30
		percent = 6
		`)

	person := Person{}
	toml.Unmarshal(document, &person)
	fmt.Printf("doc: %#v\n", person)
}

Thanks,
Will

pelletier:
Looks like the patch #236 fixed this issue, closing. Feel free to reopen if it's not the case!

PR #236, opened 2018-08-16
bugfix: panic when type unmatch between toml and struct

jayi:
When type unmatch between toml and struct, will panic.

Example:
```go
type Postgres struct {
	Id int64
}
type Config struct {
	Postgres Postgres
}

doc := []byte(`
[Postgres]
Id = 1e3
`)

config := Config{}
toml.Unmarshal(doc, &config)
fmt.Println("id=", config.Postgres.Id)
```

Panic:

```
panic: reflect: call of reflect.Value.Int on float64 Value [recovered]
	panic: reflect: call of reflect.Value.Int on float64 Value
```

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -525,7 +525,8 @@ func (d *Decoder) valueFromToml(mtype reflect.Type, tval interface{}) (reflect.V
 			if !val.Type().ConvertibleTo(mtype) {
 				return reflect.ValueOf(nil), fmt.Errorf("Can't convert %v(%T) to %v", tval, tval, mtype.String())
 			}
-			if val.Convert(mtype).Uint() < 0 {
+
+			if val.Convert(reflect.TypeOf(int(1))).Int() < 0 {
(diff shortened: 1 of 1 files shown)


pelletier:
Thanks for your PR! Looks like it's also gonna help with https://github.com/pelletier/go-toml/issues/219! I pulled your changes and tried with the example you gave, but I'm getting `config.Postgres.Id` equal to `0`. That doesn't seem like the expected output?

jayi:
```
id= 1000
```
This is the output I got, it seems ok?

pelletier:
Interesting. With

```go
type Postgres struct {
	Id int64
}
type Config struct {
	Postgres Postgres
}

doc := []byte(`
[Postgres]
Id = 1e3
`)

config := Config{}
toml.Unmarshal(doc, &config)

fmt.Println("id=", config.Postgres.Id)
```

I get `id= 0`. Which version of go are you using?

jayi:
1.9.2

pelletier:
Thanks! Sorry took so long to merge that. I was actually doing something silly, the patch works well.

jayi:
Thank you!

PR #237, opened 2018-08-22
Add Parse Type of Time.Duration Friendly, Fix Some Error Message(This is my first pull request ^_^)

Ainiroad:

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -12,7 +12,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"time"
 )
 
 var dateRegexp *regexp.Regexp
--- a/lexer.go
+++ b/lexer.go
@@ -574,7 +574,7 @@ func (l *tomlLexer) lexInsideTableArrayKey() tomlLexStateFn {
 			l.emit(tokenDoubleRightBracket)
 			return l.lexVoid
 		case '[':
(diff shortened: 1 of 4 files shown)


Ainiroad:
Ainiroad
Add Parse Type of Time.Duration Friendly

Ainiroad:
Ainiroad
Add Parse Type of Time.Duration Friendly,

pelletier:
Hello! Thanks for your PR :) I don't see a duration type in the TOML spec. Do you mind elaborating on the feature?

Ainiroad:
Thanks for your response :). Time.Duration is not in the TOML spec. But it's a common configuration types in golang. Consider the ease of use of this library, I add this feature.

pelletier:
I like the idea of the feature.

One thing I am concerned about is: when using `Load*(...)`, the current implementation would always create `Time.Duration` objects when parsing a string with the right formatting.  There would be no way to enable/disable that feature, which would create confusion if you don't know about it, or would be annoying to deal with if you do in fact want strings that look like "15m".

What I would be happy with, is to make it an opt-in flag on the `Encoder` and `Decoder`, and only un/marshal with this logic fields that are of type `time.Duration`.

Does that make sense?

pelletier:
Closing because https://github.com/pelletier/go-toml/pull/248 has been merged.

Issue #224, opened 2018-04-19
unexpected token type in inline table when using numerical key


ibrokemypie:
``
numbers = { 1 = true, 2 = true}
``
When creating an inline table as above, I get the following message
``(10, 18): unexpected token type in inline table: "1"``
however if I change the key name to one or "one" it works fine

pelletier:
Thank you for the bug report! Seems like bare keys starting with a number aren't implemented properly, even though completely legal.

ibrokemypie:
Any plans to fix this issue, or anything I can do to help?

pelletier:
Sorry for the delay. I have given that a shot a couple times. It turns out it is a non-trivial issues, that seems to require a significant amount of untangling in the lexer. I don't have much time to sit down and work on that at the moment, but I would gladly review and merge a patch!

PR #239, opened 2018-09-01
fix inlineTable parse when key is integer or string

xiehuc:
#224 

i test 

```
table =  {2018=1,2019=2}
table = {"2018"=1,"2019"=1}
```

could parse successfully to map[string]int

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -338,7 +338,7 @@ Loop:
 		case tokenRightCurlyBrace:
 			p.getToken()
 			break Loop
-		case tokenKey:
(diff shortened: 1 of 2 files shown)


pelletier:
Thank you very much for the patch! Sorry for getting back to you so late. Seems like `go fmt` is not passing. Do you mind adding a test for the parser to make sure it works as expected please? :)

xiehuc:
ok, i'd try add it when sunday

pelletier:
Awesome! Thank you!

PR #240, opened 2018-09-15
check for golang 1.11

akkumar:
ensure compatibility for golang 1.11

golang 1.11 has been released recently. Add it to travis to ensure future compatibility as well.


pelletier:
Thank you for the patch!

Issue #244, opened 2018-09-30
Gocritic 3.3 warnings: ifElseChain


KosToZyB:
Found using: https://go-critic.github.io/overview#ifElseChain-ref
```
$GOPATH/src/github.com/pelletier/go-toml/lexer.go:306:3: ifElseChain: should rewrite if-else to switch statement
$GOPATH/src/github.com/pelletier/go-toml/lexer.go:687:3: ifElseChain: should rewrite if-else to switch statement
```

PR #245, opened 2018-09-30
fixed ifElseChain warnings

KosToZyB:
fixed: https://github.com/pelletier/go-toml/issues/244

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -303,7 +303,8 @@ func (l *tomlLexer) lexKey() tomlLexStateFn {
 	growingString := ""
 
 	for r := l.peek(); isKeyChar(r) || r == '\n' || r == '\r'; r = l.peek() {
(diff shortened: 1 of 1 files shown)


lexer.go
```diff
@@ -693,19 +695,19 @@ func (l *tomlLexer) lexNumber() tomlLexStateFn {
...
 			}
-		} else if isDigit(next) {
+		case isDigit(next):
 			digitSeen = true
 			l.next()
-		} else if next == '_' {
+		case next == '_':
 			l.next()
-		} else {
+		default:
```
pelletier: I think this changes the effect of the `break` statement below, which is supposed to break out of the loop, but not breaks out of the `switch`

KosToZyB: Yes you right. returned the previous code block

PR #247, opened 2018-10-23
Allow to marshal pointer to struct and map

asenyshyn:
Have nothing much to add here. I need to be able to marshal pointer to struct or map, thus I extended validation in Encoder marshal method. As far as I can see further call of valueToTree supports both pointer and map.


asenyshyn:
No problem, thank you for review.

PR #256, opened 2018-12-19
add Delete function to tree

zenmoto:
**Issue:** none filed

This PR simply adds a 'Delete' function to the tree so that keys can be removed programatically.


pelletier:
Looks good! Do you mind adding a `DeletePath(keys []string)` (akin to `GetPath` and `SetPath`) to keep the API uniform?

Issue #216, opened 2018-02-12
commented:"true" doesn't work for non-primitive types


alankm:
Using the 'commented:"true"' tag, non-primitive types such as arrays, maps, and structs do not get commented out.

I've had a look around the source code and found that it is not trivial to implement this functionality, but I may be willing to give it a go. Was this behaviour omitted by design?

pelletier:
Good catch, thanks! It is indeed a bug – or likely a part of the implementation that was forgotten. Commenting out a non-primitive type should comment out itself and all the nodes below it.

jbsmith7741:
Just run into this issue when needing to generate a commented slice.

Issue #255, opened 2018-12-11
Keep fields order when Marshaling struct to file


troian:
In continue to few comments in #252 
This request might be dupe of #225

Sometimes we need to generate config for user (or just print existing one). There are many purpose for  this. From generating default one to debug, curiosity etc.
Usually config field are grouped to simplify context and when marshaling now all fields gets intermixed.

pelletier:
Thanks for creating this! Not quite a dup of #225, as I think this asks to enforce a specific order, while the present issue is "conserve whatever order there is".

Question: where would you expect a new field to inserted?

troian:
I think there could be an option field specifying order, `OrderPreserve`, `OrderAlphabetical`, `OrderRandom`

If order set to preserve
Old struct
```
type Obj struct {
    Val1 string `toml:"val1" comment:"val1"`
    Val3 string `toml:"val3" comment:"val3"`
}
```
resulting file
```
# val1
val1 = ""

# val3
val3 = ""
```

Update to struct
```
type Obj struct {
    Val1 string `toml:"val1" comment:"val1"`
    Val2 string `toml:"val2" comment:"val2"`
    Val3 string `toml:"val3" comment:"val3"`
}
```

resulting file
```
# val1
val1 = ""

# val2
val2 = ""

# val3
val3 = ""
```

pelletier:
@troian Hi! Just merged a #266. It adds a new `Order` option to the Encoder to preserve order of struct fields when marshaling. Let me know if it solves your issue!

troian:
Cheers, will get back with feedback soon

pelletier:
@troian closing this. Feel free to re-open if you get the chance to try it and it doesn't match your expectations :)

PR #266, opened 2019-04-01
Feature/255 keep fields ordered marshal struct

brendesp:
Here is a fix for #255

Tracking the order of members as they are marshaled so they can be written in order of the structure.  Default behavior stays the same.

New func MarshalOrdered() allows a marshalOrder to be passed in.  Current options are OrderAlphabetical (default) and OrderPreserve

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -64,7 +64,7 @@ const (
 var timeType = reflect.TypeOf(time.Time{})
 var marshalerType = reflect.TypeOf(new(Marshaler)).Elem()
 
-// Check if the given marshall type maps to a Tree primitive
+// Check if the given marshal type maps to a Tree primitive
 func isPrimitive(mtype reflect.Type) bool {
 	switch mtype.Kind() {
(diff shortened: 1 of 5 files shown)


pelletier:
Thank you so much for tackling that issue!  I know it will make a lot of people happy.

Overall the code looks good. I think `writeToOrdered`  will make it easier to fix https://github.com/pelletier/go-toml/issues/216.

However, I don't think introducing a `MarshalOrdered` function is a good interface. There is already a way to tweak the marshaling operation: the `Encoder`. I am thinking something like:

```go
var buf bytes.Buffer
err := toml.NewEncoder(&buf).Order(toml.OrderPreserve).Encode(v)
```

A bit more verbose, but reuses the existing structure and allows for more flexibility, while keeping the API as small as possible. Definitely open to suggestions though!

Thank you again.

brendesp:
I implemented you suggestions.

pelletier:
Looking great! Thank you!

PR #267, opened 2019-04-07
Add basic Dockerfile for building binaries

c-ameron:
This PR adds in a basic Dockerfile to provide an easy of use the `tomll` and `tomljson` binaries.
You could probably set up an automated build for this in docker hub as well.

Please let me know if this isn't what you had in mind/there's any changes you'd like to see 😃 

For example
```
docker build -t go-toml .
docker run -v $PWD:/workdir go-toml "tomljson" "/workdir/example.toml" | jq
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+cmd/tomll/tomll
+cmd/tomljson/tomljson
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 test_program/test_program_bin
 fuzz/
+cmd/tomll/tomll
+cmd/tomljson/tomljson
+cmd/tomltestgen/tomltestgen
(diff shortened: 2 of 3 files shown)


pelletier:
Hi @c-ameron, thank you for your PR! Interesting idea of providing a Docker image containing the tooling. I just assumed that people just used `go install` whenever they wanted to use `tomll` or `tomljson`. Do you mind expanding a bit on the use-case? Just want to get the rationale right, always careful about adding new stuff that needs to be supported. :)

c-ameron:
Yeah sure thing. The benefit being that you don't need go or anything installed to run `tomll` or `tomljson`. As long as you have docker, you're fine. `yq`, a yaml parser has a similar usecase
https://github.com/mikefarah/yq#run-with-docker
https://hub.docker.com/r/mikefarah/yq

You can then setup automated builds in docker hub to have the latest version too.

pelletier:
I think that's a good idea!

Merged this with https://github.com/pelletier/go-toml/pull/269 (added build automation + README instructions).

Thanks!

Issue #150, opened 2017-04-01
Unmarshal to map


bogem:
It will be very useful for https://github.com/nicksnyder/go-i18n/pull/66 to unmarshal TOML to maps, not only to structs.

tro3:
It should be able to marshal/unmarshal a map of string to `<type>` now, where `<type>` is anything but `interface{}`.  Or do you mean at the top level?

bogem:
I just try to execute this code:
```
var v map[string]map[string]interface{}
if err := toml.Unmarshal(tomlData, v); err != nil {
    panic(err)
}
```
And I become an error `Only a pointer to struct can be unmarshaled from TOML`. I looked into Unmarshal func and in https://github.com/pelletier/go-toml/blob/master/marshal.go#L216 I saw, what `v` should be obligatorily pointer to a struct.

tro3:
A top-level of `map[string]T`, where `T` is a specific type, is not too bad to implement.  But `map[string]interface{}` (or `map[string]map[string]interface{}` is not trivial.  I won't have time to attack that for the foreseeable future.

pelletier:
@bogem would the `ToMap()` method be enough in your case? Or do you actually have a more specific structure to map to?

bogem:
Yes, it's enough for me, but we should add some extra code to assert `ToMap()`  to needed map ([like this](https://github.com/nicksnyder/go-i18n/blob/master/i18n/bundle/bundle.go#L92)). It will be more convenient to just pass our map to `Unmarshal` function ([like this](https://github.com/nicksnyder/go-i18n/blob/master/i18n/bundle/bundle.go#L163))

pelletier:
Yes that would make sense. Marking it as a future improvement.

pelletier:
Better late than never! Since #341 got merged, the following test passes:

```go
	tomlData := `
[a]
b = 1
[c]
d = "e"
`

	expected := map[string]map[string]interface{}{
		"a": map[string]interface{}{
			"b": int64(1),
		},
		"c": map[string]interface{}{
			"d": "e",
		},
	}
	
	var v map[string]map[string]interface{}
	if err := Unmarshal([]byte(tomlData), &v); err != nil {
		panic(err)
	}
	if !reflect.DeepEqual(expected, v) {
		t.Errorf("Bad unmarshal: expected %v, got %v", expected, v)
	}
```

I it closes this issue. If the latest master version does not work for you feel free to reopen.

PR #273, opened 2019-04-16
Allow unmarshaling to top level maps

oschwald:

This is related to #150, but that seems to be for map[string]interface{}, which this PR does not attempt to address.

**Explanation of what this pull request does.**

This allows unmarshaling to a top-level map in addition to a struct.

**More detailed description of the decisions being made and the reasons why (if the patch is non-trivial).**

We were porting code over from the BurntSushi TOML parser and ran into this limitation. The implementation is trivial. I am happy to make changes.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -515,7 +515,7 @@ func (d *Decoder) SetTagName(v string) *Decoder {
 func (d *Decoder) unmarshal(v interface{}) error {
 	mtype := reflect.TypeOf(v)
 	if mtype.Kind() != reflect.Ptr {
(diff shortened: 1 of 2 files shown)


pelletier:
Thank you for your pull request! Added a couple tests and merged.

pelletier:
Feel free to submit provide additional patches and/or record issues as you transition from BurntSushi TOML. If you don't mind me asking, what are your motivations to do so?

oschwald:
Thank you! We have found that the BurntSushi parser panics on quite a few error conditions. Some of these are fixed in various PRs, but the project seems much less actively maintained. I think BurntSushi primary does Rust nowadays.

PR #276, opened 2019-04-29
Handle other key types when unmarshaling to map

oschwald:
Previously, this would fail with:

```
panic: reflect.Value.SetMapIndex: value of type string is not assignable to type toml.letter [recovered]
	panic: reflect.Value.SetMapIndex: value of type string is not assignable to type toml.letter
```


pelletier:
Good catch! Thank you!

PR #281, opened 2019-05-25
fix #279 add possibility to handle anonymous structs

### LINKED ISSUE #279: Unexpected behavior of decoding to nested anonym struct
**Describe the bug**
I tried to extend some struct to parse full toml-file and return just part of it.

**To Reproduce**
```toml
some = "value X"
another = "value Y"
```

```go
type A struct {
	Some string `toml:"some"`
}

type B struct {
	A
	Another string `toml:"another"`
}

var b B
toml.NewDecoder(r).Decode(&cnf)
fmt.Println(b)
```

**Expected behavior**
```
{{value X} value Y}
```
For example the `encoding/json` behavior: https://play.golang.org/p/x8kaSVpuqh3

**Versions**
 - go-toml: v1.3.0
 - go: 1.11.5
 - operating system: macOS

**Additional context**
Where it's happening: https://github.com/kamilsk/lift/blob/2a8adcf60311d217be784f172699d83a2031f48d/internal/config/app.go#L69-L77
TOML file: https://github.com/kamilsk/lift/blob/2a8adcf60311d217be784f172699d83a2031f48d/testdata/app.toml

kamilsk:
**Issue:** #279 

Handle anonymous struct. Add controversial case to test to solve it later.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -604,6 +604,15 @@ func (d *Decoder) valueFromTree(mtype reflect.Type, tval *Tree) (reflect.Value, 
 					}
 					mval.Field(i).Set(reflect.ValueOf(val))
 				}
+
+				// save the old behavior above and try to check anonymous structs
+				if !found && opts.defaultValue == "" && mtypef.Anonymous && mtypef.Type.Kind() == reflect.Struct {
+					v, err := d.valueFromTree(mtypef.Type, tval)
+					if err != nil {
+						return v, err
+					}
+					mval.Field(i).Set(v)
+				}
 			}
 		}
 	case reflect.Map:
--- a/marshal_test.go
+++ b/marshal_test.go
@@ -1417,3 +1417,46 @@ func TestUnmarshalDefaultFailureUnsupported(t *testing.T) {
 		t.Fatal("should error")
 	}
 }
+
+func TestUnmarshalNestedAnonymousStructs(t *testing.T) {
+	type Nested struct {
+		Value string `toml:"nested_field"`
+	}
+	type Deep struct {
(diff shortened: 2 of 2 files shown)


kamilsk:
works for me https://github.com/kamilsk/lift/commit/9afe77a85b7866a0522d0dd4c3d85b2e51ce2efc#diff-d6ad193953b17054f326552f4e2136be

pelletier:
Thank you very much for the patch!

Not sure I understand, how is the second test controversial?

kamilsk:
> Not sure I understand, how is the second test controversial?

https://github.com/pelletier/go-toml/blob/dba45d427ff48cfb9bcf633db3a8e43b7364e261/marshal.go#L552-L557

The current logic tries to pick up different keys. This may cause an error

https://github.com/pelletier/go-toml/blob/dba45d427ff48cfb9bcf633db3a8e43b7364e261/marshal.go#L566-L569

in the following scenario 

https://github.com/pelletier/go-toml/blob/dba45d427ff48cfb9bcf633db3a8e43b7364e261/marshal_test.go#L1445-L1454

For example, this is `json` behavior: https://play.golang.org/p/Y2-IqgIEUI8. I don't know how to fix it and don't break backwards compatibility.

pelletier:
Ah right. Yes, that's not ideal, but I'm happy this patch is in to at least provide partial support for people who need it. I am [rewriting the unmarshal code](https://github.com/pelletier/go-toml/tree/unmarshal-rewrite) from the ground up to behave like JSON, including for this use-case. In the meantime, this is a good-enough stopgap.

Issue #283, opened 2019-06-11
toml.Unmarshal overrides origin values even if not export field


For-ACGN:
**Describe the bug**
toml.Unmarshal overrides origin values even if not export field

**To Reproduce**
``````go
package main

import (
	"fmt"
	"log"

	"github.com/pelletier/go-toml"
)

type Ti interface {
	Print()
}

type tii struct {
	A string
}

func (this *tii) Print() {
	fmt.Print(this.A)
}

type S struct {
	A  string
	ti Ti
	tp *tii
}

func main() {
	ti := new(tii)
	tp := new(tii)
	s := &S{A: "asd", ti: ti, tp: tp}
	b, err := toml.Marshal(s)
	if err != nil {
		log.Fatalln(err)
	}
	fmt.Printf("%p %p %p \n", &s.A, s.ti, s.tp)
	toml.Unmarshal(b, s)
	fmt.Printf("%p %p %p \n", &s.A, s.ti, s.tp)
	ti.Print()
	tp.Print()
	s.ti.Print()
	s.tp.Print()
}
``````
print
``````
0xc0000884e0 0xc00005e4c0 0xc00005e4d0
0xc0000884e0 %!p(<nil>) 0x0
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x18 pc=0x4eea54]

goroutine 1 [running]:
main.main()
        E:/Project/toml bug/toml.go:41 +0x4a4
exit status 2

``````

**Expected behavior**
not set nil

**Versions**
 - go-toml: v1.4.0 and https://github.com/pelletier/go-toml/commit/dba45d427ff48cfb9bcf633db3a8e43b7364e261
 - go: version 1.12.5
 - operating system: Windows

pelletier:
Thanks for the bug report! Looks like #284 will fix that problem as soon as it's merged (it does work for this specific example).

pelletier:
@For-ACGN https://github.com/pelletier/go-toml/pull/284 has been merged. Do you mind giving this another shot to make sure the bug is fixed for you?

For-ACGN:
@pelletier @roberth-k 
Oh! It fix this bug and works right. Thank you! 
[https://github.com/For-ACGN/go-toml-test](url)

Issue #252, opened 2018-11-23
toml.Unmarshal overrides origin values if key is omitted in the file


troian:
**Describe the bug**
When unmarshaling to struct that already have some default value toml.Unmarshal overwrites it
when key is not present in the file

**To Reproduce**
Steps to reproduce the behavior. Including TOML files.
```go

package main

import(
	"github.com/pelletier/go-toml"
)

type config struct {
	Val1 string `toml:"val1"`
	Val2 string `toml:"val2"`
}

var configFile = []byte(
	`
val1 = "test1"
`)

func main() {
	cfg := &config {
		Val2: "test2",
	}

	err := toml.Unmarshal(configFile, cfg)
	if err != nil {
		panic(err.Error())
	}

	if cfg.Val2 != "test2" {
		panic("Val2 overwritten")
	}
}
```

**Expected behavior**
Should remain default value if no key is present in the file

**Versions**
 - go-toml: b56e1b2
 - go: version 1.11.2
 - operating system: macOS

fopina:
👍 just like json in stdlib https://play.golang.org/p/-iiZnh5uhy

moorereason:
@fopina,
Wrong.  `B` is retained by the `encoding/json` package.

moorereason:
I apologize.  I misunderstood your comment.  I thought you meant that go-toml was acting like the stdlib.

pelletier:
I took a quick stab at it. Not as easy as I thought it would be. The root cause is that when unmarshalling, [the whole object is replaced](https://github.com/pelletier/go-toml/blob/master/marshal.go#L462.), instead of being merged. My long-term plan to rewrite the parser inside-out to build the unmarshalled result as it runs would fix it, but just can't find the time to work on it.

In the meantime, one option would be to turn that `Set` operation into a merge operation when processing structs and maps. Not elegant but should get the job done.

troian:
Btw, another "issue" (I'm not sure it's actually issue) but might be related to parser as well
When marshaling object to stream keys order are getting into completely reordered. Is that related too?

pelletier:
[...]

PR #284, opened 2019-07-07
Merge struct fields in Unmarshal

roberth-k:
- Addresses https://github.com/pelletier/go-toml/issues/283.
- Partially addresses https://github.com/pelletier/go-toml/issues/252.

When unmarshalling (nested) structs, pass the original reflect.Value down to `valueFromTree()` and `valueFromToml()`. When that struct already exists (i.e. it's a value, or a non-nil pointer), set its fields directly instead of overwriting the struct with a blank copy.

The main effect of this change is that the library will no longer re-initialise the fields that were not available in the original TOML tree (including un-exported fields).

Setting this up as a draft request at first, as the changes have so far worked for what I needed them for, but it feels like the code has been bolted on, and may have introduced breakage somewhere. I'm open to guidance in making it merge-worthy.


pelletier:
Sorry it took a while to get back to you.

I like it! I think it's absolutely the right level of stopgap before the unmarshaller is rewritten (using a method similar to `encoding/json`).

Happy to merge this when you feel ready.

PR #292, opened 2019-08-24
Fixed #289: Float64 trucating

### LINKED ISSUE #289: Float64 trucating
**Describe the bug**
Float64 values are truncated to float 32 on Marshaling.

**To Reproduce**
Serialize a float 64 with enough precision.
ex. 
// value with too much precision for float 32
val float64 = 1.24675324675
//outputs after marshal (float 32 value)
1.2467532

**Expected behavior**
Should honor float 64 precision.

**Versions**
latest

**Additional context**
tomlValueStringRepresentation is the culprit, it does a float parse to 32 bit.  I tweaked it, but that adds resolution errors to the 32 bit tests, so figured I'd fill a report.  Let me know if this is by design, I'm assuming it isn't.

msAlcantara:
Fixed #289

This PR change the bitSize of FormatFloat to string of 32 to 64.  Maybe should be useful add a case statement to FormatFloat of float32 to string, but I don't implement because I was in doubt.

### DIFF
--- a/marshal_test.go
+++ b/marshal_test.go
@@ -55,6 +55,9 @@ Ystrlist = ["Howdy","Hey There"]
   String2 = "Three"
 `)
 
+var testMarshalFloat64 = []byte(`Value = 1.24675324675
+`)
+
 func TestBasicMarshal(t *testing.T) {
 	result, err := Marshal(basicTestData)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


marshal.go
```diff
@@ -422,7 +422,7 @@ func (e *Encoder) valueToToml(mtype reflect.Type, mval reflect.Value) (interface
 		case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
 			return mval.Uint(), nil
 		case reflect.Float32, reflect.Float64:
-			return mval.Float(), nil
+			return strconv.ParseFloat(fmt.Sprintf("%v", mval.Interface()), 64)
```
msAlcantara: This change was necessary because `mval.Float() `convert a reflect.Value of float32 to float64 with more decimal places of original data. I don't know if this is the best solution, I'm wating for reviews

Example:

> [12.3,45.6,78.9]

 was converted to 

> [12.300000190734863,45.599998474121094,78.9000015258789]

pelletier: Even though I think this is still technically correct, it is probably counter intuitive.

Your change looks good, but I would replace it to only apply the logic on the float32 case:

```go
case reflect.Float32:
	return strconv.ParseFloat(strconv.FormatFloat(float64(mval.Interface().(float32)), 'f', -1, 32), 64)
case reflect.Float64:
	return mval.Float(), nil
```

Maybe there is a better way, but this is the limit of my knowledge of floats handling in go :(

Issue #280, opened 2019-05-24
JSON to TOML tool


honkiko:
cmd/tomljson does convert toml to json. It's also very useful to have cmd/jsontoml.

pelletier:
Good idea! Thank you for the suggestion!

gdhameeja:
Hey, I'd like to work on this. Please point me in the right direction.

gdhameeja:
I wrote the code and tested it manually, it works. Will write a test and submit a PR soon.

pelletier:
Hi! Thank you for your interest in helping out! The goal of this issue is to write another command `jsontoml`, similar to `tomljson` ([code](https://github.com/pelletier/go-toml/tree/master/cmd/tomljson)).

This program is expected to read some JSON on stdin, and emit an equivalent TOML representation on stdout.

Completion checklist:

- [ ] Unit tests
- [ ] Binary copied in the [`dockerfile`](https://github.com/pelletier/go-toml/blob/master/Dockerfile)
- [ ] Mention of the tool in the [README.md](https://github.com/pelletier/go-toml/blob/master/README.md)
- [ ] Tests execution added to [.travis.yml](https://github.com/pelletier/go-toml/blob/master/.travis.yml#L21)
- [ ] Tests execution added to [.circleci/config.yaml)(https://github.com/pelletier/go-toml/blob/master/.circleci/config.yml#L66-L68)
- [ ] Tests execution added to [appveyor.yml](https://github.com/pelletier/go-toml/blob/master/appveyor.yml#L33)

I apologize for all the changes required to the various testing services. I'm hoping to clean all that up with Github actions.

Lastly, here's a the [CONTRIBUTING.md](https://github.com/pelletier/go-toml/blob/master/CONTRIBUTING.md#code-changes) document, if you haven't seen it yet.

Let me know if anything is unclear!

PR #296, opened 2019-08-29
Tool `jsontoml` is added according to ticket #280

gdhameeja:
Added tool jsontoml and tests.

**Issue:** https://github.com/pelletier/go-toml/issues/280
The tool `jsontoml` is very similar to `tomljson`
It uses json.Unmarshal to convert read json to map and then
converts the map to tree using `toml.TreeFromMap`.
Then this tree is converted to toml using `tree.toTomlString()`

The numbers when taken as input from json get converted to float64
because of how `json.Unmarshal()` converts all json numbers to float.


gdhameeja:
Awaiting review.

.circleci/config.yml
```diff
@@ -66,6 +66,10 @@ jobs:
                 test_name: "tomljson"
                 module: "github.com/pelletier/go-toml/cmd/tomljson"
                 allow_fail: <<parameters.allow_fail>>
+	    - run_test:
```
pelletier: Those tabs will need to be converted to spaces, to match the rest of the file.

Dockerfile
```diff
@@ -8,3 +8,4 @@ RUN go install ./...
 FROM scratch
 COPY --from=builder /go/bin/tomll /usr/bin/tomll
 COPY --from=builder /go/bin/tomljson /usr/bin/tomljson
+COPY --from=builder /go/bin/jsontoml /usr/bin/jsontoml
```
pelletier: Missing end-of-line character here :)

cmd/jsontoml/main.go
```diff
@@ -0,0 +1,71 @@
...
+	os.Exit(processMain(flag.Args(), os.Stdin, os.Stdout, os.Stderr))
+}
+
+func processMain(files []string, defaultInput io.Reader, output io.Writer, errorOutput io.Writer) int {
+	// read from stdin and print to stdout
+	inputReader := defaultInput
+
+	if len(files) > 0 {
+		var err error
+		inputReader, err = os.Open(files[0])
```
pelletier: I don't think the file is closed anywhere. Maybe add a `defer inputReader.Close()` after the error check?

cmd/jsontoml/main.go
```diff
@@ -0,0 +1,71 @@
...
+	return 0
+}
+
+func printError(err error, output io.Writer) {
+	io.WriteString(output, err.Error()+"\n")
+}
+
+func reader(r io.Reader) (string, error) {
+	jsonMap := make(map[string]interface{})
+	jsonBytes, err :=  ioutil.ReadAll(r); if err != nil {
```
pelletier: To be consistent with the rest of the codebase, error checking should be on its own line.

cmd/jsontoml/main.go
```diff
@@ -0,0 +1,71 @@
...
+		return "", err
+	}
+	json.Unmarshal(jsonBytes, &jsonMap)
```
pelletier: `Unmarshal` may return an error, so it needs to be checked.

cmd/jsontoml/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if output != expected {
+		t.Errorf("incorrect %s: \n%sexpected %s: \n%s", name, output, name, expected)
+		t.Log([]rune(output))
+		t.Log([]rune(expected))
+	}
+}
+
+func expectProcessMainResults(t *testing.T, input string, args []string, exitCode int, expectedOutput string, expectedError string) {
+	inputReader := strings.NewReader(input)
+	outputBuffer := new(bytes.Buffer)
```
pelletier: You should just be able to write

```go
var outputBuffer bytes.Buffer
```

as a `bytes.Buffer` needs no initialization.

.circleci/config.yml
gdhameeja: Yeah, I suppose go fmt will fix that?

Dockerfile
gdhameeja: Done.

cmd/jsontoml/main.go
gdhameeja: There seems to be no `Close()` defined on io.Reader type.

cmd/jsontoml/main.go
gdhameeja: Done

cmd/jsontoml/main.go
gdhameeja: Done

cmd/jsontoml/main_test.go
gdhameeja: This doesn't work. This is the error I get:
```
github.com/pelletier/go-toml/cmd/jsontoml/main_test.go:28:27: cannot use outputBuffer (type bytes.Buffer) as type io.Writer in argument to processMain:
	bytes.Buffer does not implement io.Writer (Write method has pointer receiver)
github.com/pelletier/go-toml/cmd/jsontoml/main_test.go:28:27: cannot use errorBuffer (type bytes.Buffer) as type io.Writer in argument to processMain:
	bytes.Buffer does not implement io.Writer (Write method has pointer receiver)
github.com/pelletier/go-toml/cmd/jsontoml/main_test.go:30:26: cannot use outputBuffer (type bytes.Buffer) as type *bytes.Buffer in argument to expectBufferEquality
github.com/pelletier/go-toml/cmd/jsontoml/main_test.go:31:26: cannot use errorBuffer (type bytes.Buffer) as type *bytes.Buffer in argument to expectBufferEquality
FAIL	github.com/pelletier/go-toml/cmd/jsontoml [build failed]
```

I got away with fixing `expectBufferEquality` by passing the addresses like &outputBuffer. But for some reason new(bytest.Buffer) works as a `io.Writer` type and var outputBuffer bytest.Buffer doesn't. Also this code was referred from the existing test for `tomljson`

.circleci/config.yml
pelletier: Nope, `go fmt` just runs on Go code, not YAML files.

cmd/jsontoml/main.go
pelletier: `os.Open` returns a `*File`, which has a `Close()` method.

gdhameeja:
Apologies, I don't understand why the build is failing. I did run `go fmt` before submitting changes.

cmd/jsontoml/main_test.go
pelletier: No worries. The reason why it fails is because the methods of the `io.Writer` interface are implemented on the pointer type (`*bytes.Buffer`), which `new` creates. You would indeed need to pass around the pointer (`&outputBuffer`).

Either way is fine with him. That wasn't a request for change, more like a FYI.

pelletier:
Did you run `go fmt` in the `cmd/jsontoml` directory?

cmd/jsontoml/main_test.go
gdhameeja: Great. Acknowledged.

cmd/jsontoml/main.go
gdhameeja: Done.

gdhameeja:
Hey, I ran go fmt on cmd/jsontoml dir. Also closed the file pointer. All tests seem to be passing. Please let me know if any further fixes are required. This will be my first open source PR with golang. I'd be very happy to see this merged. Thank you for the guidance.

pelletier:
Merged! Thank you for building that tool!

Issue #282, opened 2019-05-25
Inline table keys not fully supported


2shortplanks:
**Describe the bug**

According to the [spec](https://github.com/toml-lang/toml/blob/v0.5.0/versions/en/toml-v0.5.0.md#user-content-inline-table) in an inline table:
>  Key/value pairs take the same form as key/value pairs in standard tables

However, this library complains on anything other than something starting with a literal A-Z-a-z

**To Reproduce**

```
$ echo 'foo = { -bar => "buz"}' | ./tomljson
(1, 9): unexpected token type in inline table: Error
$ echo 'foo = { "whatever" = "buz"}' | ./tomljson
(1, 10): unexpected token type in inline table: String
echo 'foo = { _no = "buz"}' | ./tomljson
(1, 9): unexpected token type in inline table: Error
```

**Expected behavior**

```
$ echo 'foo = { -bar = "buz"}' | ./tomljson
{
  "foo": {
    "bar": "buz"
  }
}
$ echo 'foo = { "whatever" = "buz"}' | ./tomljson
{
  "foo": {
    "whatever": "buz"
  }
}
$ echo 'foo = { _no = "buz"}' | ./tomljson
{
  "foo": {
    "_no": "buz"
  }
}
```

**Versions**

```
mark@vagrant-mark:~/go/src/github.com/pelletier/go-toml/cmd/tomljson [master] $
728039f679cbcd4f6a54e080d2219a4c4928c546
mark@vagrant-mark:~/go/src/github.com/pelletier/go-toml/cmd/tomljson [master] $ go version
go version go1.11.5 linux/amd64
```

**Additional context**

Nope

pelletier:
Good catch! Thank you.

jonathanlloyd:
I'd like to have a go at this is that's OK :smile:

pelletier:
Please! Feel free to open a draft PR if you’d like early feedback too. Thank you for your interest in go-toml :)

PR #311, opened 2019-10-20
Fix key parsing in line tables

jonathanlloyd:
**Issue:** https://github.com/pelletier/go-toml/issues/282

A bug was reported that indicated that inline tables did not fully support bare keys:
$ echo 'foo = { -bar => "buz"}' | ./tomljson
(1, 9): unexpected token type in inline table: Error

$ echo 'foo = { "whatever" = "buz"}' | ./tomljson
(1, 10): unexpected token type in inline table: String

echo 'foo = { _no = "buz"}' | ./tomljson
(1, 9): unexpected token type in inline table: Error

This change makes a couple of tweaks to to allow for all key variants in inline tables

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -247,13 +247,13 @@ func (l *tomlLexer) lexRvalue() tomlLexStateFn {
 func (l *tomlLexer) lexLeftCurlyBrace() tomlLexStateFn {
 	l.next()
 	l.emit(tokenLeftCurlyBrace)
-	return l.lexRvalue
+	return l.lexVoid
 }
 
 func (l *tomlLexer) lexRightCurlyBrace() tomlLexStateFn {
 	l.next()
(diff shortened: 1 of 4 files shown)


jonathanlloyd:
Happy Hacktoberfest :jack_o_lantern: 
I'm entirely sure I'm using the void lexer state appropriately here, but it seemed like inline tables count as a new lexing context and as such are no longer rvalues?

This does cause a change in behaviour, inline tables that start with a comma now get an error from the lexer rather than the parser. However, this is the same error raised when parsing keys normally so maybe that's appropriate?

pelletier:
Looks good to me. Great work, thank you for fixing this issue!

Pushed a commit to your branch to remove the `debug.test` file, which I assumed got committed by mistake.

Issue #274, opened 2019-04-28
Default values in nested structs not used


rekby:
Default values doesn't apply for inner struct if section name doesn't present for unmarshal.

```golang
package main

import (
	"fmt"

	"github.com/pelletier/go-toml"
)

type configType struct {
	V1    int `default:"123"`
	Inner conigTypeInner
}

type conigTypeInner struct {
	V2 int `default:"456"`
}

func main() {
	var config configType
	err := toml.Unmarshal([]byte(""), &config)
	if err != nil {
		panic(err)
	}
	result, err := toml.Marshal(&config)
	if err != nil {
		panic(err)
	}
	fmt.Println(string(result))
}
```
```
Output:
V1 = 123

[Inner]
  V2 = 0
```

But if source contains section name - is ok
```golang
package main

import (
	"fmt"

	"github.com/pelletier/go-toml"
)

type configType struct {
	V1    int `default:"123"`
	Inner conigTypeInner
}

type conigTypeInner struct {
	V2 int `default:"456"`
}

func main() {
	var config configType
	err := toml.Unmarshal([]byte("[Inner]"), &config)
	if err != nil {
		panic(err)
	}
	result, err := toml.Marshal(&config)
	if err != nil {
		panic(err)
	}
	fmt.Println(string(result))
}
```

```
Output:
V1 = 123

[Inner]
  V2 = 456
```

pelletier:
Thank you for the bug report!

Top of my head, I think it does not recursively look at defaults (just checks for the `default` tag, which the `Inner` field do not have).

pelletier:
This patch should have fixed it: https://github.com/pelletier/go-toml/pull/326

If you are still experimenting issues with the latest master, feel free to reopen!

PR #326, opened 2020-01-07
Support default values for non-anonymous structs

nbedos:
**Issue:** https://github.com/pelletier/go-toml/issues/274

This adds support for default values on fields of non-anonymous structs. I'm not sure why there was a check for the struct being anonymous prior to calling `valueFromTree`? I don't see a reason to exclude non-anonymous struct.


pelletier:
Thank you for the patch! And sorry for the late reply. I don't recall why that check was there. Tests not breaking is a good enough indication for me that it was not mandatory :)

Issue #321, opened 2019-10-31
Unexpected token type in inline table


falfaroc:
**Describe the bug**
When loading a TOML file that contains an inline table with keys mapping to values that are inline tables, then the commas separating the keys are viewed as a key. 

**To Reproduce**
Toml: 
[config.toml.txt](https://github.com/pelletier/go-toml/files/3795712/config.toml.txt)
more specifically:
`out = {block0 = {x = 99, y = 100}, block1 = {p = 999, q = 1000}}`

Use the following in the go file:
```
conf, err := toml.LoadFile("config.toml")
if err != nil {
		fmt.Println("Error ", err.Error())
```

**Expected behavior**
When executing `go run main.go`, the output will be `Error  (24, 42): unexpected token type in inline table: keys cannot contain , character`

**Versions**
 - go-toml: version v1.6.0 (903d945)
 - go: version 1.13.3
 - operating system: Linux

pelletier:
Good catch. Thanks for bringing this up!

PR #334, opened 2020-03-10
Fix issue #321

AllenX2018:
**Issue:** #321 

This pr fix issue #321. The lexer should not transition its state to `lexVoid` when it meets a right curly brace. Instead, it should transition to `lexRvalue`, which can handle both simple inline table and nested inline table case.


pelletier:
Great catch! Thank you for the patch.

Issue #331, opened 2020-02-26
Support unmarshal nested interface type


jinleiw:
I think it's better to support nested Interface, For example:
```
type Conf struct {
     IntVal int
     InterVal interface{}
}

type NestedStruct struct {
    Name string
    Age int
}
var config Conf

config.InterVal = &NestedStruct{}

decoder.decode(&config)
```

pelletier:
Nested structs should be supported. Do you mind trying with name and age as public attributes (capitalized names)?

jinleiw:
Sorry, spelling mistakes. Yes, they are public attributes:

```
type NestedStruct struct {
    Name string
    Age int
}
```

jinleiw:
Could you help me review the commit “75121fd”, if ok, then I open up a pull request. Thanks

PR #335, opened 2020-03-12
Add support for nested interface{} (#331)

jinleiw:


pelletier:
Looks good to me! Thank you for the patch.

PR #336, opened 2020-03-13
fix issue#216

### LINKED ISSUE #216: commented:"true" doesn't work for non-primitive types

AllenX2018:
**Issue:** #216 
This pr fix #216. If a non-primitive(like struct, map) is commented, all of its sub kv pairs should be commented out too. Besides, I notice that If a multiline string or multiline array is commented, go-toml just commented out the first line. This pr also fix that.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -302,7 +302,7 @@ func (e *Encoder) marshal(v interface{}) ([]byte, error) {
 	}
 
 	var buf bytes.Buffer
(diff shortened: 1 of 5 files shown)


AllenX2018:
@pelletier Any thoughts?

PR #340, opened 2020-03-19
fix issue #339

### LINKED ISSUE #339: Unexpected result when nested struct
**Describe the bug**
For nested struct，if there are attributes with same name, parse result is not expected.

**To Reproduce**
```
type Server struct {
    Name string `toml:"name" default:"noname"`
    Inter2 Inter2 `toml:"inter2"`
}

type Inter2 struct {
    Name         string
    Age          int
    InterStruct2 InterStruct2
}

type InterStruct2 struct {
    Test string
    Name string
    Age  int
}
var server Server

err := toml.Unmarshal([]byte(`name = "123"

[inter2]
	name = "inter2"
	age = 222`), &server)
```

**Expected behavior**
Expected value of server:
```
{Name:"123", Inter2:{Name:"inter2", Age:222, InterStruct2:{Test:"", Name:"", Age:0}}}
```

but got:
```
{Name:"123", Inter2:{Name:"inter2", Age:222, InterStruct2:{Test:"", Name:"inter2", Age:222}}}
```

**Versions**
 - go-toml: master
 - go: 1.13.8
 - operating system: Windows

AllenX2018:
**Issue:** #339

This pr try to fix #339. And this bug seems to be introduced by #281.

### DIFF
--- a/marshal_test.go
+++ b/marshal_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/BurntSushi/toml"
 	"io/ioutil"
 	"os"
 	"reflect"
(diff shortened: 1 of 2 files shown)


marshal_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/BurntSushi/toml"
```
pelletier: I think this tests the wrong library :)

AllenX2018: Oops...Why would I get that wrong :(

PR #341, opened 2020-03-24
add interface type support

AllenX2018:
This pr introduces the support of interface{} masharling/unmarshaling. For Marshal, nil interface{} field will not be encoded, while non-nil interface{} field will be encoded referring its real type and value. For Unmarshal, if unmarshal a `Table` to a nil interface{}, a `map[string]interface{}` will be created to store its kv pairs. If unmarshal a `Table` to a non-nil interface{}, a struct referring the interface's real type will be created and try to decode the `Table`'s kv pairs to its fields.


AllenX2018:
I'm not sure if the test cases are good enough to cover most of the use case. Pls review and tell  me if you have any thoughts about  it @pelletier

PR #348, opened 2020-03-31
Support default for slice, but only one default value

jinleiw:
e.g.

```
type sliceWithDefault struct {
	Bools  []bool    `toml:"bool" default:"false"`
	Strs   []string  `toml:"name" default:"test"`
	Ints   []int     `toml:"int" default:"100"`
	Int64s []int64   `toml:"int64" default:"10000"`
	Floats []float64 `toml:"float" default:"100.111"`
}

var test sliceWithDefault
err := Unmarshal([]byte(``), &test)

// {[false] [test] [100] [10000] [100.111]}
log.Println(test)
```

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -629,6 +629,39 @@ func (d *Decoder) valueFromTree(mtype reflect.Type, tval *Tree, mval1 *reflect.V
 						if err != nil {
 							return mval.Field(i), err
 						}
+					// For slice
+					case reflect.Slice:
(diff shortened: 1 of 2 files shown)


pelletier:
Thank you for the patch. Can you elaborate a bit on the use-case for this? Supporting default for a arrays with exactly one element seems very specific.

pelletier:
Closing, as there has been no justification since march last year.

PR #353, opened 2020-04-09
dont't panic when marshal from nil or unmarshal to nil interface or pointer

AllenX2018:
This pr avoid go-toml panic when calling `Marshal(v interface{})` or `Unmarshal(data []byte, v interface{})` with nil or nil-pointer as `v`, just return an error in these cases.


pelletier:
Good idea, thank you for taking care of this!

Issue #354, opened 2020-04-11
Multiline strings can't decode tab characters


ajeetdsouza:
**Describe the bug**
Attempting to unmarshal a multiline string with a tab character throws the following error: 

```
unescaped control character U+0009
```

**To Reproduce**

```go
type Test struct {
	Inner string `multiline:"true"`
}

test := Test{Inner: "\t"}
buf, err := toml.Marshal(test)
if err != nil {
	log.Fatal(err)
}
err = toml.Unmarshal(buf, &test) // this throws an error
if err != nil {
	log.Fatal(err)
}
```

**Expected behavior**
The tab character should be decoded as expected.

**Versions**
 - go-toml: v1.7.0
 - go: 1.13.6
 - operating system: Linux (Fedora 31)

PR #355, opened 2020-04-14
fix unmarshal error with horizontal tab in multiline string

AllenX2018:
**Issue:** #354 

Horizon tab should be accepted when the lexer read multiline string from toml files.


pelletier:
From the spec:

> Any Unicode character may be used except those that must be escaped: backslash and the control characters other than tab, line feed, and carriage return 

Looks like we missed one :sweat_smile: 

Thank you for the patch!

Issue #360, opened 2020-04-15
Support leading zeroes in exponent parts of floats


pelletier:
From TOML v1.0.0-rc1 release notes:

> Leading zeroes in exponent parts of floats are permitted.

RiyaJohn:
I modified your Readme example to take a float value, like **4e-08 or 4e-0008 or 4e-8** And all of them printed **4e-08** as output for float64 type, not really sure what is the issue

pelletier:
That's great news! I added this issues because this requirement came out of the new TOML spec version. I wanted to make sure we 1. make sure it works (and update go-toml if not) and 2. is covered by tests.

PR #363, opened 2020-04-16
Add float to test case to check leading zeroes in exponent parts

RiyaJohn:
**Issue:** add link to pelletier/go-toml issue here #360

Add float to example.toml and example-crlf.toml files so that we have test case to check leading zeroes in exponent parts are supported


RiyaJohn:
@pelletier All checks passed with latest commit, not sure why PR is still stuck at waiting state

pelletier:
Sorry I haven’t had time to look at PRs recently. Thank you for taking care of this!

Issue #358, opened 2020-04-15
Allow raw tab characters in basic strings and multi-line basic strings


pelletier:
From TOML v1.0.0-rc1 release notes:

> Allow raw tab characters in basic strings and multi-line basic strings.

Seems like https://github.com/pelletier/go-toml/pull/355 already took care of the multi-line case. Need to make sure the tab character is also supported in basic strings (and tested).

pelletier:
Fixed in #364

PR #364, opened 2020-04-22
add support for tab in basic string value and quoted key

AllenX2018:
**Issue:** #358 

From [TOML v1.0.0-rc1](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v1.0.0-rc.1.md):

> Basic strings are surrounded by quotation marks. Any Unicode character may be used except those that must be escaped: quotation mark, backslash, and the control characters other than **tab** (U+0000 to U+0008, U+000A to U+001F, U+007F).

> Quoted keys follow the exact **same rules as** either **basic strings** or literal strings and allow you to use a much broader set of key names.

So the raw tab character should be allowed in quoted key, basic string as well as multiline string.

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -313,7 +313,7 @@ func (l *tomlLexer) lexKey() tomlLexStateFn {
 	for r := l.peek(); isKeyChar(r) || r == '\n' || r == '\r'; r = l.peek() {
 		if r == '"' {
 			l.next()
-			str, err := l.lexStringAsString(`"`, false, true, false)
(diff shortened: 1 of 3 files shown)


AllenX2018:
Not sure why codecov reports a coverage decrease.

pelletier:
I think it's complaining because the branch `if err != nil` after the `lexStringAsString` call is not tested. I don't think it was before so good for me.  Patch looks good. Thank you for taking care of it!

PR #365, opened 2020-04-22
fix adjacent quotation marks marshaling in multiline string

AllenX2018:
From the [spec](https://github.com/toml-lang/toml#string):

> str5 = """Here are three quotation marks: ""\\"."""
str6 = """Here are fifteen quotation marks: ""\\"""\\"""\\"""\\"""\\"."""

The adjacent quotation marks which make the multiline string closed early should be escaped when marshaling.


pelletier:
Good catch, thanks for the patch!

PR #368, opened 2020-04-24
marshal: do not encode embedded structs as sub-table

Oncilla:
Currently, the marshalling code encodes the embedded structs as sub-tables.
This is a bit unexpected, as it differs from what encoding/json does in
that case: https://play.golang.org/p/KDPaGtrijV1

This PR adapts the encoder to behave like encoding/json.
Fields in an embedded struct are promoted to the top level table.
In case the embedded struct is named in the tag, it will still
encode as a sub-table.

### DIFF
--- a/doc_test.go
+++ b/doc_test.go
@@ -5,6 +5,7 @@ package toml_test
 import (
 	"fmt"
 	"log"
+	"os"
 
 	toml "github.com/pelletier/go-toml"
 )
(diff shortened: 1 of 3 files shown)


Oncilla:
Hi, please let me know if you agree with this approach.
If so, I will be happy to update the docs accordingly.

In case we cannot do this, because it is a breaking change, we could also add another option to the tag. Something like `toml:",embed"`.

pelletier:
Great catch!  It is a breaking change, but I see it like a bugfix (it was not supposed to do this in the first place).

If you wouldn't mind, I'd love to this is as the default behavior, but maybe with an option on Encoder to be able to revert to he previous behavior in case people depend on it (I think a per-field flag is overkill).

Oncilla:
Alright. I updated the code with a `PromoteAnonymous` method on the encoder.
I struggled a bit with naming it, if you have a better suggestion, let me know.

I also changed the behavior on handling duplicate keys. It is now in line with encoding/json.
I.e., for duplicate keys, the regular field shadows the field from the anonymous struct.

This should now be ready to review.

pelletier:
Name looks good, patch is great! Thanks a lot for taking care of this.

PR #372, opened 2020-04-25
unmarshal: add strict mode

Oncilla:
This PR adds strict mode to the Decoder. It can be enabled with the
`Strict` method.

In the strict mode, the decoder fails if any fields that were part
of the input do not have a corresponding field in the struct.

The decoder only modifies the visitor state in case strict mode is enabled.
Thus, not forcing unnecessary allocations, when not needed.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -512,6 +512,8 @@ type Decoder struct {
 	tval *Tree
 	encOpts
 	tagName string
+	strict  bool
+	visitor visitorState
 }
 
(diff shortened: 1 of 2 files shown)


pelletier:
Merged in https://github.com/pelletier/go-toml/commit/d3c92c5999876ef37f03850546c93238a7841d5f (just added a test case for successful parse in strict mode). Thank you!

Oncilla:
awesome thanks :+1:

PR #374, opened 2020-04-25
marshal: support encoding.TextMarshaler

### LINKED ISSUE #373: marshal: support encoding.TextMarshaler
Most go libraries that handle marshaling/unmarshaling support `encoding.TextMarshaler` when marshaling.

Similar to how types that implement the `toml.Marshaler` interface, I would like types that implement
the `encoding.TextMarshaler` to be supported.
If a type implements both `encoding.TextMarshaler` and `toml.Marshaler`. The `MarshalToml` method should take precedence.

If that is a feature you want to integrate, I would be happy to implement it.

Oncilla:
With this PR the encoder now supports encoding.TextMarshaler.
Additionally, a bug is fixed, where the encoder does not notice a pointer
field that implements the toml.Marshaler interface.

Time types need to be special cased, because they implement the encoding.TextMarshaler
interface. Otherwise, the previous behavior cannot be maintained.

fixes #373

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -2,6 +2,7 @@ package toml
 
 import (
 	"bytes"
+	"encoding"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


pelletier:
Looks good to me, thank you for taking care of it!

Oncilla:
Thank you for the quick merges. Very appreciated :+1:

PR #375, opened 2020-04-25
unmarshal: support encoding.TextUnmarshaler

### LINKED ISSUE #329: allow custom struct unmarshal with UnmarshalText
I'm always frustrated when I want to support my struct, e.g. custom struct.

Use `encoding.TextUnmarshaler`, we can implement `UnmarshalText`.

Universal unmarshal solutions.

**Additional context**
[BurntSushi/toml](https://github.com/BurntSushi/toml#using-the-encodingtextunmarshaler-interface)

Oncilla:
This PR adds support for decoding fields of primitive types that implement
encoding.TextUnmarshaler by calling the custom method.

Fields in anonymous structs are not supported at this point.

fixes #329

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -1001,7 +1001,9 @@ func (d *Decoder) unwrapPointer(mtype reflect.Type, tval interface{}, mval1 *ref
 }
 
 func (d *Decoder) unmarshalText(tval interface{}, mval reflect.Value) error {
-	var s string
+	var buf bytes.Buffer
+	fmt.Fprint(&buf, tval)
+	return callTextUnmarshaler(mval, buf.Bytes())
 	switch val := tval.(type) {
 	case string:
 		s = val
--- a/marshal.go
+++ b/marshal.go
@@ -71,6 +71,7 @@ const (
 var timeType = reflect.TypeOf(time.Time{})
 var marshalerType = reflect.TypeOf(new(Marshaler)).Elem()
 var textMarshalerType = reflect.TypeOf(new(encoding.TextMarshaler)).Elem()
+var textUnmarshalerType = reflect.TypeOf(new(encoding.TextUnmarshaler)).Elem()
 var localDateType = reflect.TypeOf(LocalDate{})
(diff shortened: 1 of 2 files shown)


Oncilla:
@pelletier I rebased ontop of master.
I don't really know how to improve the code coverage.

One miss I can see is from:
```
func (d *Decoder) unmarshalText(tval interface{}, mval reflect.Value) error { 
	...
	default:
		return fmt.Errorf("unspported type: %t", val)
	...
```

But that should never executed, if the parser does not have a bug. Should I just put a panic there?
I think writing a test that actually triggers this error path is pretty brittle and does not really add value.

I did not figure out where the second miss comes from.

marshal.go
```diff
@@ -983,6 +1000,23 @@ func (d *Decoder) unwrapPointer(mtype reflect.Type, tval interface{}, mval1 *ref
...
+	case string:
+		s = val
+	case bool:
+		s = fmt.Sprintf("%v", val)
+	case int64:
+		s = fmt.Sprintf("%d", val)
+	case float64:
+		s = fmt.Sprintf("%f", val)
+	default:
+		return fmt.Errorf("unspported type: %t", val)
```
lmb, line 1015: I think you meant `%T`?

marshal.go
```diff
@@ -983,6 +1000,23 @@ func (d *Decoder) unwrapPointer(mtype reflect.Type, tval interface{}, mval1 *ref
 	return mval, nil
 }
 
+func (d *Decoder) unmarshalText(tval interface{}, mval reflect.Value) error {
+	var s string
```
lmb, line 1004: I think you could replace this with
```suggestion
	var buf bytes.Buffer
	fmt.Fprint(&buf, tval)
	return callTextUnmarshaler(mval, buf.Bytes())
```

marshal.go
Oncilla: No longer needed with your other suggestion

pelletier:
Looks good to me! Thanks for taking care of it. Sorry it took me a while to get to it!

Issue #357, opened 2020-04-15
Allow heterogenous values in arrays


pelletier:
From TOML v1.0.0-rc1 release notes:

> Allow heterogenous values in arrays.

I think we already support that but need to make sure we actually do and is supported by tests.

PR #376, opened 2020-04-26
add support of mixed-type array

AllenX2018:
**Issue:** #357 

This pr adds support of mixed-type array as the spec(v1.0.0) describes:

> Values of different types may be mixed.
> ```toml
> # Mixed-type arrays are allowed
> numbers = [ 0.1, 0.2, 0.5, 1, 2, 5 ]
> contributors = [
>  "Foo Bar <foo@example.com>",
>   { name = "Baz Qux", email = "bazqux@example.com", url = "https://example.com/bazqux" }
> ]
>```

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -395,7 +395,7 @@ func (e *Encoder) valueToTree(mtype reflect.Type, mval reflect.Value) (*Tree, er
 				return nil, err
 			}
 			if e.quoteMapKeys {
-				keyStr, err := tomlValueStringRepresentation(key.String(), "", "", e.arraysOneElementPerLine)
+				keyStr, err := tomlValueStringRepresentation(key.String(), "", "", OrderPreserve, e.arraysOneElementPerLine)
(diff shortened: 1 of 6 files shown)


pelletier:
Thank you for taking care of this, I apologize for the delay in review. I think the PR could use a test for the new check in `valueToOtherSlice` as it's part of the new behavior.

AllenX2018:
I think `valueToOtherSlice`'s changes has been covered by `TestMarshalMixedTypeArray()`. 

```go
if mtype.Elem().Kind() == reflect.Interface {
	return nil, fmt.Errorf("marshal can't handle []interface{}")
}
```

I delete these lines to support `[]interface` type when marshaling and `TestMarshalMixedTypeArray()` has add a test for this type. Is there any other lines/paths/changes that need to be checked?

pelletier:
Ha my bad, looks like I misread the diff. All good, thank you for calling this out!

PR #377, opened 2020-04-27
Better support for homogenous lists

### LINKED ISSUE #323: Better support of lists
A very frequent use case in TOML is using a list of strings:

```toml
my_list = [ "one", "two" ]
```

Of course in go code this would likely be handled as a slice of strings:

```go
var myList = []string
```

For this standard use case I would expect I could write something like

```go
config, err := toml.Load("a = [ 'one', 'two' ]")
if err != nil {
	log.Fatal(err)
}

myList = config.Get("my_list").([]string)
```

After struggling with go complaining about `interface conversion: interface {} is []interface {}, not []string` and finally having found #33 I realized that `go-toml` doesn't support this since TOML allow multiple different types in the same array (which actually is a very strange idea).

However having to write

```go
tmp := config.Get("my_list").([]interface{})
myList = make([]string, len(tmp))
for i := range tmp {
  myList[i] = tmp[i].(string)
}
```

seems very cumbersome to me.
As in nearly all cases an array in TOML will contain only contain values of the same type, I would expect direct support for these types (mainly []string, but also other default types) to be directly accessible without that boilerplate above.

**Additional context**
An example code snippet in the [Go Playground](https://play.golang.org/p/SFccJVomfAy)

RiyaJohn:
**Issue:** #323
Closes #323 
@pelletier 
I added logic of GetArray for homogenous lists as getArray func, but need your guidance:
1. Current changes fails the test cases for Unmarshal support. You wanted a separate function GetArray that users can directly use, but that will mean reusing GetPath() logic again in GetArray with getArray add-on. Is that okay?
2. Can you also review the getArray logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Go library for the [TOML](https://github.com/mojombo/toml) format.
 
 This library supports TOML version
-[v0.5.0](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.5.0.md)
+[v1.0.0-rc.1](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v1.0.0-rc.1.md)
 
 [![GoDoc](https://godoc.org/github.com/pelletier/go-toml?status.svg)](http://godoc.org/github.com/pelletier/go-toml)
 [![license](https://img.shields.io/github/license/pelletier/go-toml.svg)](https://github.com/pelletier/go-toml/blob/master/LICENSE)
@@ -18,7 +18,7 @@ Go-toml provides the following features for using data parsed from TOML document
 
 * Load TOML documents from files and string data
 * Easily navigate TOML structure using Tree
-* Mashaling and unmarshaling to and from data structures
+* Marshaling and unmarshaling to and from data structures
 * Line & column position data for all parsed elements
 * [Query support similar to JSON-Path](query/)
 * Syntax errors contain line and column numbers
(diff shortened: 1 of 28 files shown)


RiyaJohn:
@pelletier Please ignore the first 3 commits, not sure why my old PR commits still show up. Files changes only has changes for this pr

pelletier:
Thank you for looking at this! I think the behavior of `GetPath` should not change (hence why I'd like a new `GetArray` function. You're right that there will be a duplicate in the way traversing the tree works. I'm ok with it at the moment (eventually we should have a generic way to walk the tree).

RiyaJohn:
@pelletier I did not add test cases for homogeneous lists of ints and floats because it picks say, int64 or int32 based on system spec. Hence, test coverage decreases I guess, because otherwise all checks have passed

x-hgg-x:
I think toml integers are always parsed to int64 and toml floats are always parsed to float64, so you might safely ignore the cases for int, int32 and float32.

[Spec](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v1.0.0-rc.1.md):
For integers: ```64 bit (signed long) range expected (−9,223,372,036,854,775,808 to 9,223,372,036,854,775,807)```

For floats: ```Floats should be implemented as IEEE 754 binary64 values```

RiyaJohn:
> I think toml integers are always parsed to int64 and toml floats are always parsed to float64, so you might safely ignore the cases for int, int32 and float32 (the tree should not contain these types).
> 
> [Spec](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v1.0.0-rc.1.md):
> For integers: `64 bit (signed long) range expected (−9,223,372,036,854,775,808 to 9,223,372,036,854,775,807)`
> 
> For floats: `Floats should be implemented as IEEE 754 binary64 values`

That's great input, thank you will make changes

PR #378, opened 2020-04-28
forbid adding keys to exist inline table

AllenX2018:
As the spec describes:

> Inline tables fully define the keys and sub-tables within them. New keys and sub-tables cannot be added to them.

> ```toml
> [product]
> type = { name = "Nail" }
> # type.edible = false  # INVALID
> ```
> Similarly, inline tables can not be used to add keys or sub-tables to an already-defined table.

> ```toml
> [product]
> type.name = "Nail"
> # type = { edible = false }  # INVALID
> ```

The inline table should not be allowed to add new keys.


pelletier:
Looks good, thanks for spotting this and fixing!

PR #381, opened 2020-04-29
Fix inline table loading errors

### LINKED ISSUE #379: Failure to load an empty inline table
**Describe the bug**
Loading produces an error:
```
loading error: (1, 8): unexpected token type in inline table: keys cannot contain } character
```

**To Reproduce**
```go
_, err := toml.Load("key = {}")
if err != nil {
	fmt.Println("loading error:", err)
}
```
**Expected behavior**
There should be no error according to [the TOML specification](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.5.0.md#inline-table): 

<pre>Within the braces, <b>zero</b> or more comma separated key/value pairs may appear.</pre>

**Versions**
 - go: 1.14
 - operating system: Linux

**Additional context**
None.

### LINKED ISSUE #380: Failure to load an inline table with '-' as the 1st character in a key after a comma
**Describe the bug**
Loading produces an error:
```
loading error: (1, 21): unexpected token type in inline table: no digit in that number
```

**To Reproduce**
```go
_, err := toml.Load("key = { -123 = 'v', -key = 'v' }")
if err != nil {
	fmt.Println("loading error:", err)
}
```
**Expected behavior**
There should be no error. 

**Versions**
 - go: 1.14
 - operating system: Linux

**Additional context**
The following code works:
```go
_, err := toml.Load("key = { -key = 'v', -123 = 'v' }")
if err != nil {
	fmt.Println("loading error:", err)
}
```

x-hgg-x:
This PR fixes somes issues when parsing inline tables.

Fixes #379.
Fixes #380.

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -14,7 +14,7 @@ import (
 	"strings"
 )
 
-var dateRegexp, bareKeyWithEqualRegexp *regexp.Regexp
+var dateRegexp *regexp.Regexp
 
 // Define state functions
 type tomlLexStateFn func() tomlLexStateFn
@@ -26,7 +26,7 @@ type tomlLexer struct {
 	currentTokenStart int
 	currentTokenStop  int
 	tokens            []token
-	depth             int
+	brackets          []rune
 	line              int
 	col               int
 	endbufferLine     int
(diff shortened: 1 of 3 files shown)


pelletier:
Thank you for fixing this!

PR #383, opened 2020-04-30
Fix unmarshaler error when a custom marshaler function is defined

### LINKED ISSUE #382: Unmarshaling fails when a custom marshaler function is defined
**Describe the bug**
Unmarshaling produces an error:
```
unmarshal error: (1, 1): Can't convert B = 3
G = 2
R = 1
(*toml.Tree) to a tree
```

**To Reproduce**
```go
type color struct {
	R uint8
	G uint8
	B uint8
}

type element struct {
	Color color
}

func (color color) MarshalTOML() ([]byte, error) { return []byte("ok"), nil }

func main() {
	tree, _ := toml.Load(`[Color]
	R=1
	G=2
	B=3`)

	var e element
	err := tree.Unmarshal(&e)
	if err != nil {
	}
}
```
**Expected behavior**
There should be no error. 

**Versions**
 - go: 1.14
 - operating system: Linux

**Additional context**
None.

x-hgg-x:
This PR fixes an unmarshaling error when a custom marshaler function is defined.

Fixes #382.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -91,7 +91,7 @@ func isPrimitive(mtype reflect.Type) bool {
 	case reflect.String:
 		return true
 	case reflect.Struct:
-		return isTimeType(mtype) || isCustomMarshaler(mtype) || isTextMarshaler(mtype)
+		return isTimeType(mtype)
 	default:
 		return false
 	}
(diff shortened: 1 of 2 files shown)


pelletier:
Thank you for the patch! Do you mind adding a couple tests for the `reflect.Ptr` cases in the functions that check for custom marshaler? This seems to be an important path and would love to make sure the behavior there is correct.

pelletier:
Looking good, thank you for the patch!

PR #388, opened 2020-05-01
Fix overflow checking when unmarshaling

### LINKED ISSUE #387: Check for overflow when unmarshaling
**Describe the bug**
There are no overflow checks.

**To Reproduce**
```go
type color struct {
	R, G, B uint8
}

tree, _ := toml.Load(`
R = 200
G = 300
B = 400
`)

var c color
err := tree.Unmarshal(&c)
if err != nil {
}
fmt.Println(c)
```

Output:
```
{200 44 144}
```
**Expected behavior**
Overflow should produce an error. 

**Versions**
 - go: 1.14
 - operating system: Linux

**Additional context**
None.

x-hgg-x:
This PR fixes overflow checking when unmarshaling.

Fixes #387.


pelletier:
Good catch! Thank you for the patch.

PR #390, opened 2020-05-01
Prevent automatic conversion between int and float when unmarshaling

### LINKED ISSUE #389: Unmarshaling overflows when converting float64 to uint or int
**Describe the bug**
Unmarshaling overflows when converting float64 to uint or int if the struct field type doesn't correspond to TOML.

**To Reproduce**
```go
type element struct {
	Num uint
}

tree, _ := toml.Load("num = 1e300")

var e element
err := tree.Unmarshal(&e)
if err != nil {
}
```

Output:
```
unmarshal error: (1, 1): 1e+300(float64) is negative so does not fit in uint
```
**Expected behavior**
There should be a clearer error, and converting between float and int/uint should not be possible.

**Versions**
 - go-toml: version (git sha)
 - go: version
 - operating system: e.g. macOS, Windows, Linux

**Additional context**
None.

x-hgg-x:
This PR prevents automatic conversion between int and float when unmarshaling.

Fixes #389.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -916,7 +916,7 @@ func (d *Decoder) valueFromToml(mtype reflect.Type, tval interface{}, mval1 *ref
 				}
 				return reflect.ValueOf(d), nil
 			}
-			if !val.Type().ConvertibleTo(mtype) {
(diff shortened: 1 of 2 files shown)


pelletier:
Looks good, thank you for the fix!

PR #392, opened 2020-05-01
Add more supported types for default values when unmarshaling

### LINKED ISSUE #391: Support more types for default values when unmarshaling
I would like more supported types for default values, so that the following code works:
```go
type color struct {
	R, G, B uint `default:"128"`
}

tree, _ := toml.Load("")
var c color
err := tree.Unmarshal(&c)
if err != nil {
}
```

Now it only produces this error:
```
unmarshal error: unsupported field type for default option
```
Add more supported types.

x-hgg-x:
This PR adds more supported types for default values when unmarshaling.

Fixes #391.


pelletier:
Nice quality of life improvement. Thank you for the PR!

PR #402, opened 2020-05-06
Allow spaces when using dotted keys in assignment

### LINKED ISSUE #401: Failure to load TOML data with spaces in a dotted key assignment
**Describe the bug**
Loading produces an error:
```
(1, 5): was expecting token =, but got "." instead
```

**To Reproduce**
```go
_, err := toml.Load(` S1 . 'S2' . "S3" = 3`)
if err != nil {
	fmt.Println(err)
}
```
**Expected behavior**
There should be no error according to [the TOML specification](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.5.0.md#keys): 
```
Whitespace around dot-separated parts is ignored.
```

**Versions**
 - go-toml: last version of master (9ccd9bb)
 - go: 1.14
 - operating system: Linux

**Additional context**
The following code works:
```go
_, err := toml.Load(`[ S1 . 'S2' . "S3" ]`)
if err != nil {
	fmt.Println(err)
}
```

x-hgg-x:
This PR fixes an issue where spaces are not allowed in assignment when using dotted keys.

Fixes #401.

### DIFF
--- a/lexer_test.go
+++ b/lexer_test.go
@@ -22,20 +22,20 @@ func TestValidKeyGroup(t *testing.T) {
 }
 
 func TestNestedQuotedUnicodeKeyGroup(t *testing.T) {
-	testFlow(t, `[ j . "ʞ" . 'l' ]`, []token{
+	testFlow(t, `[ j . "ʞ" . l . 'ɯ' ]`, []token{
 		{Position{1, 1}, tokenLeftBracket, "["},
-		{Position{1, 2}, tokenKeyGroup, ` j . "ʞ" . 'l' `},
-		{Position{1, 17}, tokenRightBracket, "]"},
-		{Position{1, 18}, tokenEOF, ""},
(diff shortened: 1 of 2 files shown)


lexer_test.go
```diff
@@ -22,11 +22,20 @@ func TestValidKeyGroup(t *testing.T) {
 }
 
 func TestNestedQuotedUnicodeKeyGroup(t *testing.T) {
-	testFlow(t, `[ j . "ʞ" . l ]`, []token{
+	testFlow(t, `[ j . "ʞ" . 'l' ]`, []token{
```
pelletier, line 25: Why the change in that test?  the bare `l` should be valid right?

x-hgg-x: It is for testing at the same time bare keys, normal string keys and literal string keys.

pelletier: Oh I see what you mean. In general I'm hesitant to modify existing tests to avoid inadvertent regressions, but in that case you're right it should have worked.

pelletier:
Thank you for the patch!

PR #403, opened 2020-05-06
Restore test for accidental whitespaces

x-hgg-x:
The last PR (#381) modified an important test (#319) for accidental whitespaces after a line ending backslash in multi-line string, due to the automatic trimming of trailing whitespaces feature in my VS Code editor.

I have modified the test by visually adding whitespaces to avoid this in the future.


pelletier:
Nice catch, completely missed that during the review. Thank you for bringing them back!

PR #405, opened 2020-05-07
Fix index and slice expressions for query

x-hgg-x:
This PR fixes index and slice expressions for query.

Fixes #404.
Fixes #352.

PR #409, opened 2020-05-14
Fix query test

x-hgg-x:
This PR fix a bug in a test from the last PR (#405).


AllenX2018:
Thanks for the fix but could you pls describe why the changes for `assertArrayContainsInAnyOrder` is necessary?

x-hgg-x:
assertArrayContainsInAnyOrder is no more used, so I replaced it with a version which tests the exact order for more strict testing.

Issue #418, opened 2020-06-12
Remove/Separate benchmarking dependencies


diamondburned:
I think this module, as a TOML parser, shouldn't depend on another TOML and YML parser for benchmarking.

It would probably be best if another repository could be made for benchmarks that rely on other libraries such as BurntSushi's. Otherwise, I think it's better to remove the benchmarks.

For reference, here's the [`go.mod` as of this issue][1]: 

```
module github.com/pelletier/go-toml

go 1.12

require (
	github.com/BurntSushi/toml v0.3.1 // for benchmarking
	github.com/davecgh/go-spew v1.1.1
	gopkg.in/yaml.v2 v2.3.0 // also for benchmarking
)
```

[1]: https://github.com/pelletier/go-toml/blob/44aed552fd90fee7c7be07fd89f75b98e9fcadef/go.mod#L6

pelletier:
I agree, would be better with no dependency. I'd still like to keep the benchmarks around in this repository. Wonder if we can put them in a sub-folder with their dedicated `go.mod` file.

AllenX2018:
Good idea. I will try to move them to a sub-folder like "benchmark" with a dedicated `go.mod` file

PR #420, opened 2020-06-15
Move benchmark to a separate directory

AllenX2018:
**Issue:** #418 

1. I move the benchmark test to a folder named "benchmark" as a sub-module with a dedicated `go.mod` file.
2. It seems that different modules cannot be tested in one "go test" command so I test the "BenchmarkXXX" test cases for `github.com/pelletier/go-toml` and `github.com/pelletier/go-toml/benchmark` separately and put the result in the same file for benchstat to analyse


AllenX2018:
The CI failed because there's no "benchmark" directory in the repo until this pr is merged.

PR #425, opened 2020-07-26
Fix index exception when setting empty *Tree slice

sclevine:
When `len(node) == 0`, we can't lookup `node[len(node)-1]` unless `node` includes the newly added tree.

Example: https://play.golang.org/p/voNBW4QfRxy
Note that it works if the first key is populated: https://play.golang.org/p/8Ak-pPXVyPa

I wasn't able to use `*Tree.SetPath` to construct arbitrary TOML that works when `toml.OrderPreserve` is set.
For example: https://play.golang.org/p/Lns42AVkEcT
So I'm not sure if this fix covers a supported use case, or if logic should be removed instead.

### DIFF
--- a/toml.go
+++ b/toml.go
@@ -295,7 +295,8 @@ func (t *Tree) SetPathWithOptions(keys []string, opts SetOptions, value interfac
 			// go to most recent element
 			if len(node) == 0 {
 				// create element if it does not exist
-				subtree.values[intermediateKey] = append(node, newTreeWithPosition(Position{Line: t.position.Line + i, Col: t.position.Col}))
+				node = append(node, newTreeWithPosition(Position{Line: t.position.Line + i, Col: t.position.Col}))
+				subtree.values[intermediateKey] = node
 			}
 			subtree = node[len(node)-1]
 		}


pelletier:
Thanks for the patch! I think the issue here is that `SetPath` expects to only get the leaf values, and is in charge of building the intermediate elements of the tree. Result of bad decisions when I first wrote that library :(. In this case, omitting the intermediate `SetPath` call seems to work:

```go
	tree := emptyTree()
//	tree.SetPath([]string{"test"}, []*toml.Tree{})
	tree.SetPath([]string{"test", "a"}, "b")
	out, _ := tree.ToTomlString()
	fmt.Println(out)
```

```
[test]
  a = "b"
```

---

`OrderPreserve` is meant to preserve the order of fields when marshaling a struct, not when building a tree from scratch. Can you expand a bit on the broader objective you're trying to achieve?

sclevine:
> Can you expand a bit on the broader objective you're trying to achieve?

I'm using go-toml to convert arbitrary JSON to arbitrary TOML while preserving map order (to the extent appropriate -- tables must be moved to the bottom). To accomplish this, I'm constructing the document out of `*Tree`s with manually set `Position`s and encoding with `OrderPreserve` set.

I'm no longer using `*Tree.SetPath` incrementally, so with #426, I don't actually need this change. Just thought the negative index exception might be worth a patch.

Here's the approach I'm taking. It's messy, and I'm probably using go-toml in a way it was never intended to be used:
https://github.com/sclevine/yj/blob/master/toml/encoder.go
https://github.com/sclevine/yj/blob/master/convert/toml.go#L20-L36

But except for some issues with decoding quoted keys with escaped double-quotes (first part of #427), everything is working great! And with #426, I'm no longer on a fork 😄

pelletier (approved):
Sorry I dropped the notification on your PR!

PR #426, opened 2020-07-26
Add *Tree.SetPositionPath

sclevine:
I needed to add this function to support generating arbitrary, ordered TOML from other data serialization formats in [yj](https://github.com/sclevine/yj/).

`*Tree.SetPath` does not set `Position`s correctly when other `*Tree`s are used as values. This makes the `Encoder` behave strangely when `OrderPreserve` is set: https://play.golang.org/p/Lns42AVkEcT

Additionally, `*Tree.SetPath` doesn't appear to be sufficient to generate arbitrary TOML when `*Tree` values are not used. For example, there doesn't appear to be a way to append tables to a list.

Adding this method seems to be all that's necessary to allow arbitrary, ordered TOML to be generated, which seems useful.

### DIFF
--- a/toml.go
+++ b/toml.go
@@ -213,6 +213,50 @@ func (t *Tree) GetPosition(key string) Position {
 	return t.GetPositionPath(strings.Split(key, "."))
 }
 
+// SetPositionPath sets the position of element in the tree indicated by 'keys'.
+// If keys is of length zero, the current tree position is set.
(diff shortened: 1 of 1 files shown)


pelletier:
Thanks for the patch! Impressive we had the getters for so long without the setter 😅

sclevine:
Thanks for the quick merge! 🎉

PR #433, opened 2020-09-11
Marshal into empty interface{}

### LINKED ISSUE #432: Support Unmarshal into empty interface{}
While looking at migrating [Hugo](https://github.com/gohugoio/hugo) from BurntSushi to this package, I found that this package doesn't support unmarshaling into an empty `interface{}`.  Returns an error "_only a pointer to struct or map can be unmarshaled from TOML._"

When an empty `interface{}` is passed to `toml.Unmarshal` as the `v` parameter, decode as a `map[string]interface{}`.

Stick with BurntSushi. 😞 

**Additional context**
* Desired usage:  https://play.golang.org/p/CKHM44HsRgP
* JSON stdlib does this: https://play.golang.org/p/oM0Y08WGuf5
* yaml.v3 does this: https://play.golang.org/p/IMAelUFHQ_p (yaml.v2 creates a `map[interface{}]interface{}`, but that idea was abandoned in v3)

pelletier:
Allows to marshal a TOML document into an empty `interface{}`, resulting
in a `map[string]interface{}`.

Fixes #432

cc @moorereason

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -76,6 +76,7 @@ var textUnmarshalerType = reflect.TypeOf(new(encoding.TextUnmarshaler)).Elem()
 var localDateType = reflect.TypeOf(LocalDate{})
 var localTimeType = reflect.TypeOf(LocalTime{})
 var localDateTimeType = reflect.TypeOf(LocalDateTime{})
+var mapStringInterfaceType = reflect.TypeOf(map[string]interface{}{})
 
 // Check if the given marshal type maps to a Tree primitive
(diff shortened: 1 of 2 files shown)


moorereason:
Wow.  So simple.  Works great.  Only had to change the import path to switch from BurntSushi<sup>1</sup>.  🏆 

<sup>1</sup> <i>Although, some of our marshaling tests are not expecting inline tables, but I'll investigate that issue separately.</i>

PR #444, opened 2020-09-12
toml.Unmarshaler supports leaf nodes

### LINKED ISSUE #437: Unmarshaler doesn't respect Unmarshaler interface on type mismatch
**Describe the bug**

Decoder doesn't respect Unmarshaler interface if source (toml) and destination value types are different.
For example if source type is `string` but unmarshaled onto struct which implements `Unmarshaler` interface.

I found a workaround by implementing `TextUnmarshaler` interface, but if it's an expected behavior - it should be documented.

**To Reproduce**

See demo - https://goplay.tools/snippet/Dxp0lfxX7Dy

**Expected behavior**

Instead of complaining about types, decoder should just pass value to `Unmarshaler` like `encoding/json` package does.

**Versions**
 - go-toml: [v1.8.0](https://github.com/pelletier/go-toml/releases/tag/v1.8.0)
 - go: 1.15
 - operating system: Linux

**Additional context**

N/A

pelletier:
Fixes #437

### DIFF
--- a/marshal_test.go
+++ b/marshal_test.go
@@ -3,6 +3,7 @@ package toml
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


marshal_test.go
```diff
@@ -3976,3 +3977,59 @@ func TestGithubIssue431(t *testing.T) {
...
+	cfg := &config437{}
+	cfg.HTTP.PingTimeout = durationString{time.Second}
+
+	r := strings.NewReader(src)
+	err := NewDecoder(r).Decode(cfg)
+	if err != nil {
+		t.Fatalf("unexpected errors %s", err)
+	}
+	expected := durationString{10 * time.Second}
+	if cfg.HTTP.PingTimeout != expected {
```
x1unix, line 4018: @pelletier I recommend you to take a look at [`github.com/stretchr/testify`](https://github.com/stretchr/testify) package.

It contains a set of handy tools for unit tests - [`require`](https://pkg.go.dev/github.com/stretchr/testify@v1.6.1/require?tab=doc) and [`assert`](https://pkg.go.dev/github.com/stretchr/testify@v1.6.1/assert?tab=doc)

```go
// assert.NoError is also ok, but require will stop test if condition is failed
require.NoError(t, err)
require.EqualError(t, err, "some error message")
require.Equal(t, expected, cfg.HTTP.PingTimeout)
```

pelletier: Thanks!

PR #453, opened 2020-11-09
Fix ToMap for tables in mixed-type arrays

mstetson:
ToMap doesn't convert tables to maps if they're inline tables in a mixed-type array.

```toml
a = ["foo", { bar = "baz" }]
```

Before this change, `ToMap` will return a map where the second element of the array `a` will be a `*toml.Tree` instead of a `map[string]interface{}`.


pelletier:
Awesome! Thank you for fixing this!

PR #455, opened 2020-11-19
Fix ToMap for tables in nested mixed-type arrays

mstetson:
I wasn't thorough enough in my earlier PR. #453 fixed tables in mixed arrays, but only in the top-level array, not in any nested array. This updated fix scans arrays recursively looking for tables. It supports constructs like this:

```toml
UglyValue = ["blah", [
    {foo = "bar"},
    {baz = "quux"},
]]
```

### DIFF
--- a/tomltree_write.go
+++ b/tomltree_write.go
@@ -510,27 +510,29 @@ func (t *Tree) ToMap() map[string]interface{} {
 		case *Tree:
 			result[k] = node.ToMap()
 		case *tomlValue:
-			result[k] = node.toValue()
+			result[k] = tomlValueToGo(node.value)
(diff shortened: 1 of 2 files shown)


pelletier:
Thanks! Is there a case where this branch is taken?

```
	if tv, ok := v.(*tomlValue); ok {
		v = tv.value
	}
```

If yes it's probably worth a test, if not we can get rid of the code :)

pelletier:
Merged with https://github.com/pelletier/go-toml/pull/461

Thank you!

mstetson:
Thanks! Sorry I left you hanging there.

PR #464, opened 2021-01-18
Improve default for durations

terorie:
**Summary**: This improves support for the `default` struct tag on `time.Duration`.

**Problem**

`time.Duration` is an `int64` type, so it's already supported by the default tag.
However, the default is interpreted as a number of nanoseconds, not a duration, making parsing a bit awkward.

Example for 120ms:
```go
var doc struct {
    DurationField time.Duration `default:"120000000"`
}
```

**Feature**

Go's `time` library supports a human-readable notation for durations.
Including this improves readability:

```go
var doc struct {
    DurationField time.Duration `default:"120ms"`
}
```

**Implementation notes**

To preserve backwards-compatibility, the code checks for three conditions:
* The type is `int64`
* The type name is `time.Duration`
* The last char of the provided default is not a number

If either is not met, the code falls back to parsing a number.

**Open questions:**

* Is the backwards-compatibility check overkill, or good means of preventing people from shooting themselves in the foot?
* Instead of checking `.Type().String()`, is it preferred to check for `.Type().PkgName()` and `.Type().Name()` for a tiny bit more performance?

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -830,7 +830,21 @@ func (d *Decoder) valueFromTree(mtype reflect.Type, tval *Tree, mval1 *reflect.V
 					case reflect.Int32:
 						val, err = strconv.ParseInt(opts.defaultValue, 10, 32)
 					case reflect.Int64:
-						val, err = strconv.ParseInt(opts.defaultValue, 10, 64)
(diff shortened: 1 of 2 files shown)


pelletier (approved):
Looks good, thank you for the patch!

PR #465, opened 2021-01-23
Update docs to reference pkg.go.dev

y-yagi:
godoc.org  is still alive. But requests are going to redirect to pkg.go.dev. 
This fixed to point to pkg.go.dev direct to reduce unneeded redirect.

Ref: https://blog.golang.org/godoc.org-redirect


pelletier (approved):
Thank you!

PR #467, opened 2021-01-28
fix #466: use type alias to make Tree and tomlValue props publicly av…

### LINKED ISSUE #466: make tree struct properties and tomlValue public
I'm trying to enable toml for https://github.com/mozilla/sops with this lib, but it's quite difficult not to be able to work directly with the comment/values/tomlValue in order to parse the toml.Tree into sops.TreeBranches.
I tried to work with the map ( tree.ToMap() ) but it makes things quite complicated for no reason (on top of losing the comments and positions).

Is there any reason why these types/properties were kept private?
Any possibility of making these public? or creating a new public struct type and a new function allowing the access to these infos?

vincentserpoul:
…ailable

**Issue:** #466 

Explanation of what this pull request does: allow access to private props

### DIFF
--- a/tomlpub.go
+++ b/tomlpub.go
@@ -19,6 +19,22 @@ func (ptv *PubTOMLValue) Position() Position {
 	return ptv.position
 }
 
+func (ptv *PubTOMLValue) SetValue(v interface{}) {
+	ptv.value = v
+}
+func (ptv *PubTOMLValue) SetComment(s string) {
+	ptv.comment = s
+}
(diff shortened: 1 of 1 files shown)


pelletier:
Looks good to me. Do you also need setters methods as well?

vincentserpoul:
you re right, I will need it to rebuild the tree back from the sops tree.
added!

pelletier (approved):

Issue #459, opened 2020-12-23
Import parameters type marshalOrder int


456vv:
type marshalOrder int
    func (e *Encoder) Order(ord marshalOrder) *Encoder

should be:

    type MarshalOrder int
    func (e *Encoder) Order(ord MarshalOrder) *Encoder

pelletier:
Good catch, will update!

PR #469, opened 2021-02-01
add value string representation pub function

### LINKED ISSUE #468: encoding value to toml
**Describe the bug**
I can't get a toml encoded value from an input value.

**Expected behavior**
For example, let's say I want to write into toml a []int{1, 2 ,3}, I would expect a function to take []int{1, 2 ,3} as an argument and return []byte([1, 2, 3] 

**Additional context**
Last needed function to finalize my PR for mozilla/sops
Should I just wrap func tomlValueStringRepresentation into a public function?

vincentserpoul:
fix #468

### DIFF
--- a/tomltree_writepub.go
+++ b/tomltree_writepub.go
@@ -1,0 +1,6 @@
+package toml
+
+// ValueStringRepresentation transforms an interface{} value into its toml string representation.
+func ValueStringRepresentation(v interface{}, commented string, indent string, ord MarshalOrder, arraysOneElementPerLine bool) (string, error) {
+	return tomlValueStringRepresentation(v, commented, indent, ord, arraysOneElementPerLine)
+}


pelletier:
Can you share some code that shows how this function would be used?

vincentserpoul:
mozilla/sops is used to encrypt values of json/yaml/dotenv files, and at decrypt time, it needs to transform the encrypted value into a valid string, according to the store format.
You can look at [the yaml implementation](https://github.com/mozilla/sops/blob/38b25bd449619e1d6da20e637702f7c73203aa44/stores/yaml/store.go#L193).

This function is used [here](https://github.com/mozilla/sops/blob/38b25bd449619e1d6da20e637702f7c73203aa44/cmd/sops/decrypt.go#L69).

Maybe there is a better function to use, let me know.

vincentserpoul:
For example, let's say I want to write into toml a []int{1, 2 ,3}, I would expect a function to take []int{1, 2 ,3} as an argument and return []byte("[1, 2, 3]")

pelletier:
You can actually do this by using `SetPath` and casting your value to go-toml's internal type: https://play.golang.org/p/0oUpk5zKipc

```
func setInts(t *toml.Tree, path []string, values []int) {
	interfaces := make([]interface{}, len(values))
	for i, x := range values {
		interfaces[i] = interface{}(int64(x))
	}
	t.SetPath(path, interfaces)
}
```

I realize this is less than ideal -- I am working on a proof-of-concept for a [v2](https://github.com/pelletier/go-toml/tree/v2-wip) of go-toml that wouldn't be plagued by those usability issues (among other things).

Hopefully the cast trick is good enough to unblock you -- otherwise we can maybe add some more Set* functions for appropriate types.

vincentserpoul:
I only need the value though, without the key associated to it.
That's why I was thinking the function I wrapped would work. Do you think it would not?

vincentserpoul:
And thanks a lot for your reactivity and helpfulness!

pelletier:
Oh sorry I just realized you didn't need the whole tree. Exposing this function makes sense.

Don't you need to also expose `marshalOrder` ?

vincentserpoul:
you re right, forgot about it.

I guess that's one more method/function you want to expose in your v2!

pelletier:
Actually I just realized `marshalOrder` should be exposed anyway => https://github.com/pelletier/go-toml/issues/459. Do you mind exposing it as `MarshalOrder` directly instead of a new type? I can do it later tonight otherwise.

vincentserpoul:
Or did you mean replace it entirely in the codebase?

pelletier:
Exposed it here: https://github.com/pelletier/go-toml/pull/470

vincentserpoul:
rebased

pelletier (approved):

PR #474, opened 2021-02-04
add linting with golangci-lint

vincentserpoul:
Note the config is using toml :)

I configured with the most constraints, of course you can add or remove whatever you think is useful.

It would be nice to pair it with a github action running after each push.

### DIFF
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -1,0 +1,9 @@
+[service]
+golangci-lint-version =  "1.36.0"
+
+[linters]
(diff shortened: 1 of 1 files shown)


pelletier:
Nice! Do you know if there is a way to make it report errors in-line the same way codecov does?

vincentserpoul:
I think the easiest is to use github actions: https://github.com/marketplace/actions/run-golangci-lint

PR #483, opened 2021-03-11
Remove go-spew dependency

sapphire-janrain:
Removed the go-spew dependency since it's not really necessary. There were two places it was being used.

For parser_test I recreated the same form Sdump was outputting with pure fmt flags.

This was not possible for toml_testgen_support_test without reimplementing the functionality of go-spew, so I just did a basic, non-pretty dump. Personally, I think the non-pretty dump is better than prettying it up with json.MarshalIndent (which loses typing information but does not seem necessary). The reason for this is that it's easy to copy/redirect the terminal out to a file and visually compare the flattened lines to see when they stop being the same, this is much more difficult with pretty-printing as you have to jump back and forth etc. If you don't agree, I'm happy to change it to MarshalIndent.

For the original outputs vs the ones produced by the changes in this PR (and also the json.MarshalIndent version), see [here](https://gist.github.com/sapphire-janrain/fedff592624d9a31b1a163dbdc00ed8a).


pelletier (approved):
That's fair. Forgot it was still in there. Thank you!

PR #486, opened 2021-03-29
v2: Fix scanning of float with leading zero

moorereason:


pelletier:
Good catch, thank you!

PR #492, opened 2021-04-02
Golangci lint v2 WIP

vincentserpoul:
lint for v2


jidicula:
`internal/imported_tests/unmarshal_imported_test.go` has a lot of unused variables that `golangci-lint` is flagging. Is it ok if I nuke them or are they going to be used as the v2 WIP continues?

vincentserpoul:
@pelletier I think regular merges/checks would help a lot to avoid a big painful rebase if you continue development

pelletier:
Ack, I'll look at this today!

decode.go
```diff
@@ -48,31 +48,43 @@ func parseLocalDate(b []byte) (LocalDate, error) {
...
 
 	date.Day, err = parseDecimalDigits(b[8:10])
+	if err != nil {
+		return date, err
+	}
 
 	return date, nil
 }
 
+var ErrNotDigit = errors.New("not a digit")
```
pelletier, line 62: I'd like to avoid exporting errors at the moment, as I haven't fleshed out what the API for handling errors should be.

decode.go
```diff
@@ -48,31 +48,43 @@ func parseLocalDate(b []byte) (LocalDate, error) {
...
 		v *= 10
 		v += int(c - '0')
 	}
+
 	return v, nil
 }
 
+var ErrParseDateTimeMissingInfo = errors.New("date-time missing timezone information")
+
+const dateTimeByteLen = 6
```
pelletier, line 81: Cool with naming the const, but I'd rather see it defined close to where it is used.

decode.go
```diff
@@ -123,166 +135,234 @@ func parseDateTime(b []byte) (time.Time, error) {
...
 	hasBefore := false
+
 	for idx, r := range value {
 		if r == '_' {
 			if !hasBefore || idx+1 >= len(value) {
 				// can't end with an underscore
 				return errInvalidUnderscore
 			}
 		}
+
```
pelletier, line 339: Out of curiosity, are those newlines enforced by golangci-lint or is this personal preference?

errors.go
```diff
@@ -123,50 +128,67 @@ func wrapDecodeError(document []byte, de *decodeError) error {
...
 	var beforeLines [][]byte
 
-	// Walk the document in reverse from the highlight to find previous lines
+	// Walk the document backward from the highlight to find previous lines
 	// of context.
 	rest := document[:offset]
+backward:
 	for o := len(rest) - 1; o >= 0 && len(beforeLines) <= linesAround && len(rest) > 0; {
-		if rest[o] == '\n' {
+		switch {
```
pelletier, line 147: I prefer the `if .. else if ... else` form as it does not require labeling the `for` loop to break. Is this something that can be disabled in the linter?

errors_test.go
```diff
@@ -134,21 +139,22 @@ line 3
...
-1| line1
-2| 
-3| line 2
-4| before highlighted after
- |        ~~~~~~~~~~~ 
-5| line 3
-6| 
-7| line 4
-`,
+			expected: "" +
```
pelletier, line 142: The point of the original literal string use is that examples could easily be typed and copy-pasted. I'm not keen on losing that.

internal/ast/ast.go
```diff
@@ -11,7 +11,7 @@ import (
 // it := n.Children()
 // for it.Next() {
 // 		it.Node()
-// }
+// }.
```
pelletier, line 14: Guessing those extra dots were added by an auto-fix of some sort? :D If so let's revert and disable that.

localtime.go
```diff
@@ -44,15 +44,17 @@ type LocalDate struct {
 func LocalDateOf(t time.Time) LocalDate {
 	var d LocalDate
 	d.Year, d.Month, d.Day = t.Date()
+
```
pelletier, line 47: Not found of those extra new lines.

decode.go
```diff
@@ -123,166 +135,234 @@ func parseDateTime(b []byte) (time.Time, error) {
...
+	return f, nil
 }
 
 func parseIntHex(b []byte) (int64, error) {
 	tok := string(b)
 	cleanedVal := cleanupNumberToken(tok)
-	err := hexNumberContainsInvalidUnderscore(cleanedVal)
+
+	if err := hexNumberContainsInvalidUnderscore(cleanedVal); err != nil {
```
pelletier, line 266: I would like to keep the function call and error handling separate. I find the (arguable) readability and the consistency with functions returning more than one value to outweigh the conciseness and scope of err.

pelletier (changes requested):
Thank you for working through those! There are some style changes that I don't support, but I think once tuned having the linter can help remove style discussions from pull requests and avoid mistakes like not handling errors (my bad!)

pelletier:
@jidicula unused variables might be referenced in commented-out code in those tests. They are commented out because the API for those tests isn't there yet. In those cases, I think the variables should be commented out too. However I suspect there may be some legit unused variables in those tests.

decode.go
vincentserpoul: This is my mistake, I should not have capitalized them right away, I will lowercase all the errors.

decode.go
vincentserpoul: will do, maybe even within the function actually

decode.go
vincentserpoul: golangci-lint, I won't enforce any personal opinions :).

internal/ast/ast.go
jidicula: Must be an autofixer complying with recommendation in Go's [list of code review comments](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences).

errors.go
vincentserpoul: I thin it should be. The recommendation comes from the gocritic linter and is following idiomatic go https://golang.org/doc/effective_go#switch

errors_test.go
vincentserpoul: I 100% agree with you, it's much better.
My problem is that if you leave it like this, anyone using gofmt will break the test (as it happened to me) without noticing.
Basically, it removes trailing spaces, breaking the test.
Also, I didn't take the time to check but is it normal that the test breaks without the trailing spaces?

internal/ast/ast.go
vincentserpoul: indeed, the godot linter, which wants... dots at the end of comments

localtime.go
vincentserpoul: it's the wsl linter. We can deactivate if you prefer
I think the default confs usually make sense, I broke many of my habits to fit them and honestly got use to it, it makes things kinda easier to read. It's your choice though.

From the wsl README:

WSL is a linter that enforces a very non scientific vision of how to make code more readable by enforcing empty lines at the right places.

pelletier:
FYI I ran the benchmarks between the base of this branch and tip, and benchstat is reporting some regression on the UnmarshalSimple test:

```
name                  old time/op    new time/op    delta
UnmarshalSimple/v2-8    1.02µs ± 3%    1.04µs ± 2%  +1.78%  (p=0.012 n=10+10)
ReferenceFile/v2-8      34.3µs ± 4%    35.0µs ± 5%    ~     (p=0.190 n=10+10)

name                  old speed      new speed      delta
ReferenceFile/v2-8     153MB/s ± 4%   150MB/s ± 5%    ~     (p=0.190 n=10+10)

name                  old alloc/op   new alloc/op   delta
ReferenceFile/v2-8      11.8kB ± 0%    11.8kB ± 0%    ~     (all equal)

name                  old allocs/op  new allocs/op  delta
ReferenceFile/v2-8         253 ± 0%       253 ± 0%    ~     (all equal)
```

(using `go test '-bench=.*/v2' -count=10`)

Not sure what would be causing this though.

vincentserpoul:
> FYI I ran the benchmarks between the base of this branch and tip, and benchstat is reporting some regression on the UnmarshalSimple test:
> 
> ```
> name                  old time/op    new time/op    delta
> UnmarshalSimple/v2-8    1.02µs ± 3%    1.04µs ± 2%  +1.78%  (p=0.012 n=10+10)
> ReferenceFile/v2-8      34.3µs ± 4%    35.0µs ± 5%    ~     (p=0.190 n=10+10)
> 
> name                  old speed      new speed      delta
> ReferenceFile/v2-8     153MB/s ± 4%   150MB/s ± 5%    ~     (p=0.190 n=10+10)
> 
> name                  old alloc/op   new alloc/op   delta
> ReferenceFile/v2-8      11.8kB ± 0%    11.8kB ± 0%    ~     (all equal)
> 
> name                  old allocs/op  new allocs/op  delta
> ReferenceFile/v2-8         253 ± 0%       253 ± 0%    ~     (all equal)
> ```
> 
> (using `go test '-bench=.*/v2' -count=10`)
> 
> Not sure what would be causing this though.

a few hundreds nanosecs, is it systematic? Could it be your device having a different background noise?

vincentserpoul:
Let's recap here the conclusions of these lints:
* New errors or const should not be made public
* const should be declared as close as possible as their use
* Revert all autofixes
* autofix to be done one file by one file, to be sure it's not doing dumb things
* do not use switch when it brings a label (to be confirmed)
* wsl to be deactvated (to be confirmed)
*

jidicula:
> @jidicula unused variables might be referenced in commented-out code in those tests. They are commented out because the API for those tests isn't there yet. In those cases, I think the variables should be commented out too. However I suspect there may be some legit unused variables in those tests.

Ok, I'll comment them out instead of removing them so the linter can be silenced _for now_. I'll mark them with TODO as well for further consideration when the API has been built out.

vincentserpoul:
> > @jidicula unused variables might be referenced in commented-out code in those tests. They are commented out because the API for those tests isn't there yet. In those cases, I think the variables should be commented out too. However I suspect there may be some legit unused variables in those tests.
> 
> Ok, I'll comment them out instead of removing them so the linter can be silenced _for now_. I'll mark them with TODO as well for further consideration when the API has been built out.


you can also use //nolint:deadcode

decode.go
vincentserpoul: actually, I re-ran the linter, and it advocates for this on only specific cases. So it seems I have been a bit overzealous on this one.
I will rerun it as well.

errors.go
pelletier: Given this example is called out in effective go, let's use the switch form you propose and i'll get used to it. Thanks for pointing this out!

errors_test.go
pelletier: It seems to me that the issue is the code generating the error is appending extra whitespaces. You're right: the test should pass when you trim end-of-line spaces in the test expected value.

localtime.go
pelletier: Looking at the wsl configuration, I would set `allow-assign-and-anything: true`.

pelletier:
> a few hundreds nanosecs, is it systematic? Could it be your device having a different background noise?

Re-ran it and it looked better:

```
$ benchstat bench_origin_v2.txt bench_golangci_v2.txt         
name                  old time/op    new time/op    delta
UnmarshalSimple/v2-8    1.02µs ± 3%    1.03µs ± 4%   ~     (p=0.671 n=10+10)
ReferenceFile/v2-8      34.3µs ± 4%    34.9µs ± 5%   ~     (p=0.218 n=10+10)

name                  old speed      new speed      delta
ReferenceFile/v2-8     153MB/s ± 4%   150MB/s ± 5%   ~     (p=0.218 n=10+10)

name                  old alloc/op   new alloc/op   delta
ReferenceFile/v2-8      11.8kB ± 0%    11.8kB ± 0%   ~     (all equal)

name                  old allocs/op  new allocs/op  delta
ReferenceFile/v2-8         253 ± 0%       253 ± 0%   ~     (all equal)
```

I was just flagging it to make sure we don't trigger something like unexpected escaping with the refactoring.

vincentserpoul:
I rebased on @jidicula merge, kept my previous mods on a separate branch and restarted on this one, added the conf for wsl (but there are still added spaces) and took into account your remarks about errors and const.

Let me know if that's ok for you, before I continue on the rest @pelletier

pelletier (approved):
This pr looks good to me @vincentserpoul! Would it make it easier for you if I merge it into v2-wip?

vincentserpoul:
it would be great to merge indeed, bit by bit so you can also have the config on your local, to be able to enable golangci-lint when you dev on your side as well!

PR #493, opened 2021-04-05
Address golangci-lint warnings in unmarshal_imported_test.go

jidicula:
**Issue:**#492

Following discussion with @vincentserpoul and @pelletier, this PR addresses linting issues only in `unmarshal_imported_test.go`, which are mostly unused type and var warnings. Instead of removing the unused types and vars, I've instead chosen to mark them with a TODO and `//nolint` comment to both silence the linter and mark the symbols for later review when their dependents are added to the v2 rewrite.

I've split up each warning that's addressed by commit - each commit message body also names the symbol being marked. Note that there's 1 commit handling a lint error in a different file (`tracker.go`) that I'm including in this PR as well since it's a relatively small change.


pelletier:
Thank you!

PR #510, opened 2021-04-15
Simplify scanFollows usage

moorereason:
Use static functions to avoid declaring global vars and creating more
package init costs.  This change has no negative effects on benchmarks
in my testing.

### DIFF
--- a/scanner.go
+++ b/scanner.go
@@ -2,29 +2,37 @@ package toml
 
 import "fmt"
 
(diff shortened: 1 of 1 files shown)


pelletier:
Interesting, I would have expected to see an allocation because of `string(b[:n])`, but it does not seem to be the case.

moorereason:
I was half-expecting the same thing.  Here's [a tiny playground proof](https://play.golang.org/p/1YvmcoZFGWd).

PR #512, opened 2021-04-15
decoder: strict mode

### LINKED ISSUE #500: v2: decoder: strict mode
Decoder should offer a strict mode that returns an error when some values in the input data do not have a corresponding target in the document.

pelletier:
Fixes #500


pelletier:
Realized we probably want to provide a way to access keys. Converting to draft.

PR #517, opened 2021-04-19
V2 lint marshaler

vincentserpoul:
marshaler linting fixes


vincentserpoul:
actually a bad idea to create another separate PR, I ll just force push in the other PR.
Sorry for the mess :)

PR #518, opened 2021-04-21
V2 golangci lint

vincentserpoul:
I noticed some linters trigger only in vscode, which can lead to erratic changes. Not sure if it happens on your sides.
Anyway, I tried to split this one a bit better

### DIFF
--- a/localtime_test.go
+++ b/localtime_test.go
@@ -293,9 +293,9 @@ func TestDateTimeToString(t *testing.T) {
 		dateTime  LocalDateTime
 		roundTrip bool // ParseLocalDateTime(str).String() == str?
 	}{
-		{"2016-03-22T13:26:33", LocalDateTime{LocalDate{2016, 0o3, 22}, LocalTime{13, 26, 33, 0}}, true},
-		{"2016-03-22T13:26:33.000000600", LocalDateTime{LocalDate{2016, 0o3, 22}, LocalTime{13, 26, 33, 600}}, true},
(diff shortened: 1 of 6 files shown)


localtime_test.go
```diff
@@ -293,9 +293,9 @@ func TestDateTimeToString(t *testing.T) {
 		dateTime  LocalDateTime
 		roundTrip bool // ParseLocalDateTime(str).String() == str?
 	}{
-		{"2016-03-22T13:26:33", LocalDateTime{LocalDate{2016, 03, 22}, LocalTime{13, 26, 33, 0}}, true},
-		{"2016-03-22T13:26:33.000000600", LocalDateTime{LocalDate{2016, 03, 22}, LocalTime{13, 26, 33, 600}}, true},
-		{"2016-03-22t13:26:33", LocalDateTime{LocalDate{2016, 03, 22}, LocalTime{13, 26, 33, 0}}, false},
+		{"2016-03-22T13:26:33", LocalDateTime{LocalDate{2016, 0o3, 22}, LocalTime{13, 26, 33, 0}}, true},
```
pelletier, line 296: can we just put `3`? an octal literal seems weird :D

pelletier:
> I noticed some linters trigger only in vscode, which can lead to erratic changes. Not sure if it happens on your sides.
I'm yet to set this up on my end. I'll let you know when I do!

localtime_test.go
vincentserpoul: yep, not sure how this came up... it seems to be gofumpt, but not sure. Luckily it was caught by you.

pelletier: In Go when a literal number is prefixed by `0` it implies the octal notation. I guess whatever tool fixed that decided to make it explicit

PR #523, opened 2021-04-22
V2 golangci lint

vincentserpoul:
global vars will be the theme of this one


marshaler.go
```diff
@@ -643,13 +643,14 @@ func (enc *Encoder) encodeTableInline(b []byte, ctx encoderCtx, t table) ([]byte
...
 
-var (
-	errNilInterface   = errors.New("nil interface not supported")
-	textMarshalerType = reflect.TypeOf(new(encoding.TextMarshaler)).Elem()
-)
+var errNilInterface = errors.New("nil interface not supported")
+
+func textMarshalerType() reflect.Type {
+	return reflect.TypeOf(new(encoding.TextMarshaler)).Elem()
+}
```
vincentserpoul, line 651: I usually try to avoid global vars as much as possible and replace it with a function.
I'm pretty sure the compiler will inline the function and replace the call anyway and it stops from having a global var.

The only exception I usually do is for errors (no choice) and for compiler updated vars (like a version, for ex).

The benefit is to be 100% sure this var will never be overwritten
The drawback is the hackiness of it

I know this can be controversial, so debate is open!
I can bench as well if you re not convinced there is no perf cost.

marshaler.go
```diff
@@ -165,28 +165,37 @@ func (ctx *encoderCtx) isRoot() bool {
...
 func (enc *Encoder) encode(b []byte, ctx encoderCtx, v reflect.Value) ([]byte, error) {
-	//nolint:gocritic,godox
-
 	if v.Type() == timeType {
-		i := v.Interface().(time.Time)
+		i, ok := v.Interface().(time.Time)
+		if !ok {
+			return nil, errInterfaceNotTime
+		}
+
```
vincentserpoul, line 181: very unlikely to happen (maybe even impossible to have a timeType not castable into a time, but... well, in case the code around is changing, this might catch something one day.

marshaler.go
pelletier: Benchmarking with the following code, the difference is significant:

```go
func BenchmarkTextMarshalerFn(b *testing.B) {
	v := reflect.ValueOf(struct{}{})
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		v.Type().Implements(textMarshalerType())
	}
}

var textMarshalerTypeVar = reflect.TypeOf(new(encoding.TextMarshaler)).Elem()

func BenchmarkTextMarshalerVar(b *testing.B) {
	v := reflect.ValueOf(struct{}{})
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		v.Type().Implements(textMarshalerTypeVar)
	}
}
```

```
name                 time/op
TextMarshalerFn-32   53.9ns ± 2%
TextMarshalerVar-32  6.87ns ± 2%

name                 alloc/op
TextMarshalerFn-32    16.0B ± 0%
TextMarshalerVar-32   0.00B     

name                 allocs/op
TextMarshalerFn-32     1.00 ± 0%
TextMarshalerVar-32    0.00     
```

Given that there is an alloc every op, it does not seem like the compiler is about to optimize this out (haven't double checked the emitted instructions).

The benefit seem slim here, given that to be overwritten the value would need to be of the same type to pass compilation and would need to be close enough to pass the test suite. I wish Go would let us define those as const or have literal types, but unfortunately the next best thing is to compute that value once.

marshaler_test.go
```diff
@@ -342,28 +345,31 @@ type flagsSetters []struct {
...
+func allFlags() flagsSetters {
+	return flagsSetters{
+		{"arrays-multiline", (*toml.Encoder).SetArraysMultiline},
+		{"tables-inline", (*toml.Encoder).SetTablesInline},
+	}
 }
 
 func setFlags(enc *toml.Encoder, flags int) {
-	for i := 0; i < len(allFlags); i++ {
+	for i := 0; i < len(allFlags()); i++ {
```
pelletier, line 356: Function call seems to bring noise, but no real benefit.

marshaler.go
vincentserpoul: this just me shows once again that I shouldn't trust my intuition, stop talking and just bench it :).
I did bench this change in different situations, but always ended up being the same results and no alloc. I guess this one of the benefit of participating in OSS: getting exposed to more use cases :).

I'll remove the two global var commits

marshaler.go
vincentserpoul: I ll bench it and come back with that. I feel it's clearer and simpler to just have the cast with the ok, but let's see, I feel weak with my assumptions now :P

vincentserpoul:
```golang
var timeType = reflect.TypeOf(time.Time{})

func v1(v reflect.Value) time.Time {
	if v.Type() == timeType {
		return v.Interface().(time.Time)
	}

	return time.Time{}
}

func v2(v reflect.Value) time.Time {
	i, ok := v.Interface().(time.Time)
	if ok {
		return i
	}

	return time.Time{}
}
```
```golang
func BenchmarkV1(b *testing.B) {
	x := []reflect.Value{reflect.ValueOf(0), reflect.ValueOf(time.Now())}

	b.ReportAllocs()
	b.ResetTimer()

	// run the Fib function b.N times
	for n := 0; n < b.N; n++ {
		v1(x[n%2])
	}
}

func BenchmarkV2(b *testing.B) {
	x := []reflect.Value{reflect.ValueOf(0), reflect.ValueOf(time.Now())}

	b.ReportAllocs()
	b.ResetTimer()

	// run the Fib function b.N times
	for n := 0; n < b.N; n++ {
		v2(x[n%2])
	}
}
```

```
name  time/op
V1-8  9.36ns ± 1%
V2-8  6.42ns ± 2%

name  alloc/op
V1-8   0.00B     
V2-8   0.00B
```

so it seems better to simply try to cast and check if ok. It's also clearer imho.

pelletier: sounds good to me!

go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/pelletier/go-toml/v2
 
-go 1.15
+go 1.16
```
pelletier, line 3: The go version should stay at 1.15 because go-toml supports the last two versions of go. It will be bumped to 1.16 when 1.17 is released.

marshaler.go
```diff
@@ -632,9 +636,11 @@ func (enc *Encoder) encodeTableInline(b []byte, ctx encoderCtx, t table) ([]byte
 	return b, nil
 }
 
-var errNilInterface = errors.New("nil interface not supported")
-
-var textMarshalerType = reflect.TypeOf(new(encoding.TextMarshaler)).Elem()
+var (
+	errNilInterface = errors.New("nil interface not supported")
+	//nolint:gochecknoglobals
```
pelletier, line 641: I think this linter can be disabled.

go.mod
vincentserpoul: makes sense, I wasn't aware of that

marshaler.go
vincentserpoul: I hope this won't come bite later! 
commented the linter and removed traces of it

go.mod
```diff
@@ -2,7 +2,4 @@ module github.com/pelletier/go-toml/v2
 
 go 1.15
 
-require (
-	github.com/stretchr/testify v1.7.0
-	golang.org/x/perf v0.0.0-20210220033136-40a54f11e909 // indirect
```
vincentserpoul: not sure fi there is a reason for keeping this?

pelletier: not sure why it was added in the first place

pelletier: i try to limit the number of dependencies of go-toml. used to have 0 on v1, but i made an exception for testify on v2.

pelletier (approved):
Looking good!

pelletier:
Thank you!

PR #529, opened 2021-04-26
V2 golangci lint

vincentserpoul:


scanner.go
```diff
@@ -111,6 +111,7 @@ func scanWhitespace(b []byte) ([]byte, []byte) {
 	return b, b[len(b):]
 }
 
+//nolint:unparam
```
vincentserpoul, line 117: scanComment - result 0 ([]byte) is never used (unparam)
Not sure if it's temporar or permanent, so I left it here. It basically says the first result is always ignored

unmarshaler.go
```diff
@@ -90,49 +94,63 @@ func (d *decoder) arrayIndex(append bool, v reflect.Value) int {
...
-	}
 	if err == nil {
-		err = d.strict.Error(p.data)
+		return d.strict.Error(p.data)
+	}
+
+	var e *decodeError
+	if errors.As(err, &e) {
+		return wrapDecodeError(p.data, e)
 	}
```
vincentserpoul, line 110: probably the highlight of this linting session: casting errors should be done this way!

unmarshaler_test.go
```diff
@@ -1171,11 +1238,15 @@ key3 = "value3"
...
 	fmt.Println(err.Error())
-	// Output: strict mode: fields in the document are missing in the target struct
 
-	details := err.(*toml.StrictMissingError)
+	var details *toml.StrictMissingError
+	if !errors.As(err, &details) {
+		panic(fmt.Sprintf("err should have been a *toml.StrictMissingError, but got %s (%T)", err, err))
+	}
+
 	fmt.Println(details.String())
```
vincentserpoul, line 1247: Spelling mistake was preventing the complete test

unmarshaler_test.go
```diff
@@ -1171,11 +1238,15 @@ key3 = "value3"
...
-	// Ouput:
+	// Output:
```
vincentserpoul, line 1248: Spelling mistake was preventing the complete test

scanner.go
pelletier: Not using it now, but we will when implementing the Document interface, which will allow to manipulate comments.

scanner.go
```diff
@@ -100,27 +110,31 @@ func scanWhitespace(b []byte) ([]byte, []byte) {
...
 		}
 	}
-	return b, nil, nil
+
+	return b, nil
 }
 
+var errBasicLineNotTerminatedByQuote = errors.New(`basic string not terminated by "`)
+
+//nolint:godox
```
pelletier, line 137: The TODO can actually be removed

scanner.go
```diff
@@ -142,22 +156,22 @@ func scanBasicString(b []byte) ([]byte, []byte, error) {
 		}
 	}
 
-	return nil, nil, fmt.Errorf(`basic string not terminated by "`)
+	return nil, nil, errBasicLineNotTerminatedByQuote
 }
 
+//nolint:godox
```
pelletier, line 162: Same here. Those TODOs were here when those functions were `parse*` functions in the initial v2 code, but now there are `scan*` and `decode*` functions, with the latter handling validation.

unmarshaler.go
pelletier: neat!

unmarshaler.go
```diff
@@ -252,6 +285,7 @@ func (d *decoder) scopeWithArrayTable(x target, key ast.Iterator) (target, bool,
 		x, err = scopeSlice(true, x)
 	case reflect.Array:
 		x, err = d.scopeArray(true, x)
+	default:
```
pelletier, line 288: That doesn't seem useful 😄

unmarshaler_test.go
pelletier: 🤦🏻‍♂️ good catch

pelletier:
Merged it. I'll delete the TODOs when I do a full pass on them.

unmarshaler.go
vincentserpoul: indeed... just to satisfy the linter. I hesitated between nolint and this, but i guess it's less bad this way :)

PR #530, opened 2021-04-28
V2 golangci lint

vincentserpoul:
should be the last straw for the root folder.

### DIFF
--- a/targets.go
+++ b/targets.go
@@ -39,26 +39,31 @@ func (t valueTarget) get() reflect.Value {
 
 func (t valueTarget) set(v reflect.Value) error {
 	reflect.Value(t).Set(v)
+
 	return nil
 }
 
 func (t valueTarget) setString(v string) error {
 	t.get().SetString(v)
+
 	return nil
 }
 
 func (t valueTarget) setBool(v bool) error {
 	t.get().SetBool(v)
+
 	return nil
 }
 
 func (t valueTarget) setInt64(v int64) error {
 	t.get().SetInt(v)
+
 	return nil
 }
 
 func (t valueTarget) setFloat64(v float64) error {
 	t.get().SetFloat(v)
+
 	return nil
 }
 
@@ -128,6 +133,7 @@ func (t mapTarget) get() reflect.Value {
 
 func (t mapTarget) set(v reflect.Value) error {
 	t.v.SetMapIndex(t.k, v)
+
 	return nil
 }
 
@@ -184,6 +190,7 @@ func ensureValueIndexable(t target) error {
 	case reflect.Ptr:
 		if f.IsNil() {
 			ptr := reflect.New(f.Type().Elem())
+
 			err := t.set(ptr)
 			if err != nil {
 				return fmt.Errorf("ensureValueIndexable: %w", err)
@@ -191,6 +198,7 @@ func ensureValueIndexable(t target) error {
 
 			f = t.get()
 		}
+
 		return ensureValueIndexable(valueTarget(f.Elem()))
 	case reflect.Array:
 		// arrays are always initialized.
@@ -201,8 +209,10 @@ func ensureValueIndexable(t target) error {
 	return nil
 }
 
-var sliceInterfaceType = reflect.TypeOf([]interface{}{})
-var mapStringInterfaceType = reflect.TypeOf(map[string]interface{}{})
+var (
+	sliceInterfaceType     = reflect.TypeOf([]interface{}{})
+	mapStringInterfaceType = reflect.TypeOf(map[string]interface{}{})
+)
 
 func ensureMapIfInterface(x target) error {
 	v := x.get()
@@ -528,47 +538,53 @@ func (d *decoder) scopeTableTarget(shldAppend bool, t target, name string) (targ
 
 	switch x.Kind() {
 	// Kinds that need to recurse
-
 	case reflect.Interface:
 		t, err := scopeInterface(shldAppend, t)
 		if err != nil {
-			return t, false, err
+			return t, false, fmt.Errorf("scopeTableTarget: %w", err)
 		}
+
 		return d.scopeTableTarget(shldAppend, t, name)
 	case reflect.Ptr:
 		t, err := scopePtr(t)
 		if err != nil {
-			return t, false, err
+			return t, false, fmt.Errorf("scopeTableTarget: %w", err)
 		}
+
 		return d.scopeTableTarget(shldAppend, t, name)
 	case reflect.Slice:
 		t, err := scopeSlice(shldAppend, t)
 		if err != nil {
-			return t, false, err
+			return t, false, fmt.Errorf("scopeTableTarget: %w", err)
 		}
 		shldAppend = false
+
 		return d.scopeTableTarget(shldAppend, t, name)
 	case reflect.Array:
 		t, err := d.scopeArray(shldAppend, t)
 		if err != nil {
-			return t, false, err
+			return t, false, fmt.Errorf("scopeTableTarget: %w", err)
 		}
 		shldAppend = false
+
 		return d.scopeTableTarget(shldAppend, t, name)
 
 	// Terminal kinds
-
 	case reflect.Struct:
 		return scopeStruct(x, name)
 	case reflect.Map:
 		if x.IsNil() {
-			t.set(reflect.MakeMap(x.Type()))
+			err := t.set(reflect.MakeMap(x.Type()))
+			if err != nil {
+				return t, false, fmt.Errorf("scopeTableTarget: %w", err)
+			}
+
 			x = t.get()
 		}
 
 		return scopeMap(x, name)
 	default:
-		panic(fmt.Errorf("can't scope on a %s", x.Kind()))
+		panic(fmt.Sprintf("can't scope on a %s", x.Kind()))
 	}
 }
 
@@ -577,6 +593,7 @@ func scopeInterface(shldAppend bool, t target) (target, error) {
 	if err != nil {
 		return t, err
 	}
+
 	return interfaceTarget{t}, nil
 }
 
@@ -585,6 +602,7 @@ func scopePtr(t target) (target, error) {
 	if err != nil {
 		return t, err
 	}
+
 	return valueTarget(t.get().Elem()), nil
 }
 
@@ -593,7 +611,13 @@ func initPtr(t target) error {
 	if !x.IsNil() {
 		return nil
 	}
-	return t.set(reflect.New(x.Type().Elem()))
+
+	err := t.set(reflect.New(x.Type().Elem()))
+	if err != nil {
+		return fmt.Errorf("initPtr: %w", err)
+	}
+
+	return nil
 }
 
 // initInterface makes sure that the interface pointed at by the target is not
@@ -616,9 +640,10 @@ func initInterface(shldAppend bool, t target) error {
 	} else {
 		newElement = reflect.MakeMap(mapStringInterfaceType)
 	}
+
 	err := t.set(newElement)
 	if err != nil {
-		return err
+		return fmt.Errorf("initInterface: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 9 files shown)


targets.go
```diff
@@ -121,6 +153,12 @@ func (t mapTarget) setFloat64(v float64) error {
 	return t.set(reflect.ValueOf(v))
 }
 
+var (
+	errValIndexExpectingSlice  = errors.New("expecting a slice")
+	errValIndexCanNotInitSlice = errors.New("cannot initialize a slice")
```
moorereason, line 158: Casing of this var should be `...Cannot...`?  Same for a few others in this PR.

targets.go
```diff
@@ -129,159 +167,319 @@ func ensureValueIndexable(t target) error {
...
+	errSetInt64InUint64    = errors.New("negative integer does not fit in an uint64")
+	errSetInt64InUint32    = errors.New("negative integer does not fit in an uint32")
+	errSetInt64InUint32Max = errors.New("integer does not fit in an uint32")
+	errSetInt64InUint16    = errors.New("negative integer does not fit in an uint16")
+	errSetInt64InUint16Max = errors.New("integer does not fit in an uint16")
+	errSetInt64InUint8     = errors.New("negative integer does not fit in an uint8")
+	errSetInt64InUint8Max  = errors.New("integer does not fit in an uint8")
+	errSetInt64InUint      = errors.New("negative integer does not fit in an uint")
+	errSetInt64Unknown     = errors.New("does not fit in an uint")
+)
```
moorereason, line 301: I didn't notice this until just now, but...if errors are only used in one place, I'd suggest they not be declared globally.  These global vars bloat package init costs for no real reason.

targets.go
```diff
@@ -291,95 +489,111 @@ func elementAt(t target, idx int) (target, error) {
...
 		return elementAt(valueTarget(f.Elem()), idx)
 	default:
-		return nil, fmt.Errorf("cannot elementAt on a %s", f.Kind())
+		return nil, fmt.Errorf("elementAt: %w on a %s", errElementAtCannotOnUnknown, f.Kind())
 	}
 }
 
-func (d *decoder) scopeTableTarget(append bool, t target, name string) (target, bool, error) {
+//nolint:cyclop
+func (d *decoder) scopeTableTarget(shldAppend bool, t target, name string) (target, bool, error) {
```
moorereason, line 536: Nitpick: I would just spell out "should" as `shouldAppend`.

targets.go
```diff
@@ -291,95 +489,111 @@ func elementAt(t target, idx int) (target, error) {
...
 		return scopeMap(x, name)
 	default:
-		panic(fmt.Errorf("can't scope on a %s", x.Kind()))
+		panic(fmt.Sprintf("can't scope on a %s", x.Kind()))
 	}
 }
 
-func scopeInterface(append bool, t target) (target, error) {
-	err := initInterface(append, t)
+func scopeInterface(shldAppend bool, t target) (target, error) {
```
moorereason, line 591: Same

targets.go
```diff
@@ -396,13 +611,19 @@ func initPtr(t target) error {
...
+	}
+
+	return nil
 }
 
 // initInterface makes sure that the interface pointed at by the target is not
 // nil.
 // Returns the target to the initialized value of the target.
-func initInterface(append bool, t target) error {
+func initInterface(shldAppend bool, t target) error {
```
moorereason, line 626: Same

targets.go
```diff
@@ -414,54 +635,63 @@ func initInterface(append bool, t target) error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("initInterface: %w", err)
 	}
 
 	return nil
 }
 
-func scopeSlice(append bool, t target) (target, error) {
+func scopeSlice(shldAppend bool, t target) (target, error) {
```
moorereason, line 652: Same

targets.go
```diff
@@ -414,54 +635,63 @@ func initInterface(append bool, t target) error {
...
 		v = t.get()
 	}
+
 	return valueTarget(v.Index(v.Len() - 1)), nil
 }
 
-func (d *decoder) scopeArray(append bool, t target) (target, error) {
+var errScopeArrayNotEnoughSpace = errors.New("not enough space in the array")
+
+func (d *decoder) scopeArray(shldAppend bool, t target) (target, error) {
```
moorereason, line 672: Same

unmarshaler.go
```diff
@@ -81,7 +81,7 @@ type decoder struct {
 	strict strict
 }
 
-func (d *decoder) arrayIndex(append bool, v reflect.Value) int {
+func (d *decoder) arrayIndex(shldAppend bool, v reflect.Value) int {
```
moorereason, line 84: Same

moorereason (commented):
Suggesting some changes.

targets.go
vincentserpoul: I'm surprised the misspell linter actually didn't tell me anything.
English is not my mother tongue, so sometimes, this kind of things can happen :). Will correct that.

targets.go
vincentserpoul: I wish go had a better way of declaring errors...
unfortunately, if they are not placed outside of the function, then they really have no point to exist.
The goal is for other functions to be able to detect what errors was triggered. Some of these will be exported, but no way to know which ones right now..

vincentserpoul: just did the test. The impact I see for 1000 declared errors is a 200kB in the generated binary.
What else are you concerned about that I should check @moorereason ?

moorereason: I'm talking about package init costs not binary size.  For example, given:

```go
package main

import (
        _ "github.com/pelletier/go-toml/v2"
)

func main() {
        return
}
```

Show package init costs with the Go toolchain on `master`:

```
$ git checkout master
$ GODEBUG=inittrace=1 go run . 2>&1 | grep '^init.*toml'
init github.com/pelletier/go-toml/v2 @0.79 ms, 0.015 ms clock, 288 bytes, 11 allocs
```
Now show init costs of this PR:
```
$ git checkout v2-golangci-lint
$ GODEBUG=inittrace=1 go run . 2>&1 | grep '^init.*toml'
init github.com/pelletier/go-toml/v2 @0.69 ms, 0.009 ms clock, 1736 bytes, 67 allocs
```
Those are the package init costs for simply importing `go-toml` into your project.  Go will run all init logic for all packages before it gets to running your code.  For Go libraries, we generally want that to be a small as possible.

vincentserpoul: Makes complete sense!

So we can either:
* default errs outside functions and put them in if they don't need to be exported or used by other functions.
* default errs inside the functions and put them out when they need to be used/exported.

Followed by a last pass before v2 release to check which are used and which are exported.

I guess that's a @pelletier decision!

moorereason: One point of clarification:  I compared `master` to your PR, but I should have been comparing the `v2` branch...so:
```
GODEBUG=inittrace=1 go run . 2>&1 | grep '^init.*toml'
init github.com/pelletier/go-toml/v2 @1.4 ms, 0.022 ms clock, 1368 bytes, 44 allocs
```

pelletier: That's a good point. I haven't really considered the impact on initialization speed. As a rule, I favor keeping the "happy path" fast, even if it means being slower on the error path.

Let's remove them from the global init and declare them inline. We can also assume that they will not be exported. Ideally, most (all?) of them should really be `decodeError`s to provide the full context.

targets.go
```diff
@@ -71,23 +77,48 @@ func (t interfaceTarget) get() reflect.Value {
 }
 
 func (t interfaceTarget) set(v reflect.Value) error {
-	return t.x.set(v)
+	err := t.x.set(v)
+	if err != nil {
+		return fmt.Errorf("interfaceTarget set: %w", err)
```
pelletier, line 82: why wrapper all errors?

targets_test.go
```diff
@@ -31,14 +34,22 @@ func TestStructTarget_Ensure(t *testing.T) {
 			test: func(v reflect.Value, err error) {
 				assert.NoError(t, err)
 				require.False(t, v.IsNil())
-				s := v.Interface().([]string)
+
+				s, ok := v.Interface().([]string)
+				if !ok {
+					panic(fmt.Sprintf("interface %v should be castable into []string", s))
```
pelletier, line 40: The single assign form for casting will panic with a nice error message when the cast fails, no need to do it manually.

toml_testgen_support_test.go
```diff
@@ -49,74 +50,98 @@ func testgenValid(t *testing.T, input string, jsonRef string) {
 	require.Equal(t, refDoc, doc2)
 }
 
-type testGenDescNode struct {
-	Type  string
-	Value interface{}
-}
+// type testGenDescNode struct {
```
pelletier, line 53: If it's not used it can be deleted.

targets.go
```diff
@@ -129,159 +167,319 @@ func ensureValueIndexable(t target) error {
...
+func ensureMapIfInterface(x target) error {
 	v := x.get()
+
 	if v.Kind() == reflect.Interface && v.IsNil() {
 		newElement := reflect.MakeMap(mapStringInterfaceType)
-		x.set(newElement)
+
+		err := x.set(newElement)
+		if err != nil {
+			return fmt.Errorf("ensureMapIfInterface: %w", err)
```
pelletier, line 225: Good catch on the missing err handling, but I think this should panic. It's a program error if an element that was just created cannot be set.

pelletier (changes requested):
I'd like to understand the reasoning behind wrapping all errors with `fmt.Errorf("function name: %w", err)`. This seems to create a lot of noise especially since all those function names are internal.

The extra check + panic for casts should be removed.

moorereason:
See https://blog.golang.org/go1.13-errors

I agree about the internal function references. More thought needs to be put into how to do this effectively.

pelletier:
Sorry, I meant specifically wrapping errors just adding function names, basically creating a stacktrace in the error message. In general I think wrapping errors is great when it adds context meaningful to the user, which doesn't seem to be the case here.

targets_test.go
vincentserpoul: then, we probably have to add a nolint :/. There s not config as far as I can see.

vincentserpoul:
just to clarify the why, I just drop that here:

```
package main

import (
	"errors"
	"fmt"
	"log"
	"time"
)

func main() {
	if err := a(); err != nil {
		log.Printf("%v", err)
	}
}

func a() error {
	err := errors.New("int does not fit")
	if time.Now().Unix()%5 == 0 {
		return fmt.Errorf("a: %w", err)
	}

	return fmt.Errorf("%w", b())
}

func b() error {
	return fmt.Errorf("b: %w", errors.New("int does not fit"))
}
```

As far as I understand, if there is no function name, there is no way to know which function a or b triggered the error.
You can argue that all errors should be distinct, but nothing enforces that and I've been mad so many times at error messages returned from a lib with no way to know their origin, that I took the habit to add function names in there.

Maybe there is a better way, I'd be glad to learn it!

targets.go
vincentserpoul: I think it all goes down to [my comment here](https://github.com/pelletier/go-toml/pull/530#issuecomment-829783833)

vincentserpoul:
the error handling definitely needs more discussions. Should we merge this and then I can start a PR for the errors cleanup/refactor? Or you prefer to solve this here and expand it to other files in another PR?

targets.go
pelletier: Let's disable this linter.

After reading the post from that author and Dave Cheney, this is linter seems like the wrong enforcement.

targets_test.go
pelletier: Let's disable `forcetypeassert`. I'm not sure about why this would be a good idea to enforce it, maybe for programs that should never-ever panic? Not our case here.

pelletier:
Sorry for the delay. @vincentserpoul agreed, error handling needs more thoughts. I'm happy with merging this after fixing the other non-error things. We can keep the new `fmt.Errorf` for not as a way to grep through when we figure out a good way to do error handling.

targets_test.go
vincentserpoul: Don't you think it enforces contributor to consciously panic? I feel it's a safeguard that is a little noisy, but can catch some errors at some point?

vincentserpoul: I sometimes refer to the [Uber golang style guide](https://github.com/uber-go/guide/blob/master/style.md#dont-panic). Definitely not THE reference, but usually a good point of view. This is a test though, I know it doesnt really matter.

vincentserpoul:
anything I forgot in there?

pelletier:
Thank you!

Issue #450, opened 2020-10-19
Ordered-Marshal doesn't give the correct order and in certain cases duplicates the value after using SetPath


heyaWorld:
**Describe the bug**
```NewEncoder(w io.Writer).Order(OrderPreserve).Encode(v interface{})``` doesn't return the correct ordered output and in certain cases duplicates the path set by (*Tree).SetPath()

**To Reproduce**

```
    config := `
	[[inputs.cpu]]
	    collect_cpu_time = false
	
        [[inputs.apache]]
	     urls = []
	    [inputs.apache.tags]
		foo="bar"

      [[inputs.disk]]`

	// Load config, output is *Tree
	tree, _ := toml.Load(config)
	node := tree.GetPath([]string{"inputs"}).(*toml.Tree)
	for _, key := range node.Keys() {
		tree.SetPath([]string{"inputs", key, "tags", "computer"}, "mycomputerName")
		tree.SetPath([]string{"inputs", key, "tags", "origin"}, "telegraf/apache")
	}
	fmt.Println(tree)
	fmt.Println("******************************\n")

	// Convert to bytes to preserve the order of the original config
	var result bytes.Buffer
	if err := toml.NewEncoder(&result).Order(toml.OrderPreserve).Encode(tree); err != nil {
		fmt.Println(err)
	}
	fmt.Println("************original ordered toml with tags************")
	fmt.Println(string(result.Bytes()))
```
   
**Expected behavior**
```[inputs.cpu.tags] ``` should not be duplicate.  
```"collect_cpu_time = false"``` is missing under the inputs.cpu table.

**Original Ordered output after running the above code:**

```
[inputs]

  [[inputs.cpu]]

    [inputs.cpu.tags]
      computer = "mycomputerName"
      origin = "telegraf/apache"

    [inputs.cpu.tags]
      computer = "mycomputerName"
      origin = "telegraf/apache"

  [[inputs.apache]]
    urls = []

    [inputs.apache.tags]
      foo = "bar"
      computer = "mycomputerName"
      origin = "telegraf/apache"

  [[inputs.disk]]

    [inputs.disk.tags]
      computer = "mycomputerName"
      origin = "telegraf/apache"

```

**Expected ordered output:**

```
`[inputs]

  [[inputs.cpu]]
    collect_cpu_time = false
    [inputs.cpu.tags]
      computer = "mycomputerName"
      origin = "telegraf/apache"

  [[inputs.apache]]
    urls = []

[...]

PR #539, opened 2021-05-10
Return non-zero for empty trees from getTreeArrayLine

Felixoid:
**Issue:** #450

Empty trees cause overriding of any existing node with line 0. It's not fixing the mentioned issue completely, although it works 100% for the case I've faced today during playing around with the library.


Felixoid:
I'd like to add the test to cover my case as well, so I am converting it to the draft for now

pelletier:
Looks good! Thank you very much for the patch!

Issue #540, opened 2021-05-11
Is it an option an implementing `Encoder.CompactComments(bool)` method?


Felixoid:
Hello. I continue to playing around with the library.

I've found, that with the current code it's impossible to get the compact config output when comments are used. Each comment produces an additional line and it's difficult to overview the config. As a workaround, I use the following snippet, but it removes spaces between sections too:

```
out := strings.Replace(buf.String(), "\n", "", 1)
out = strings.ReplaceAll(out, "\n\n", "\n")
```

So, I have now the config like 

```
[common]

# general listener
listen = ":9090"
max-cpu = 1

# limit number of results from find query, 0=unlimited
max-metrics-in-find-answer = 0

# limit numbers of queried metrics per target in /render requests, 0 or negative = unlimited
max-metrics-per-target = 15000

# daemon will return the freed memory to the OS when it>0
memory-return-interval = "0s"

# listener to serve /debug/pprof requests. '-pprof' argument overrides it
pprof-listen = ""

# daemon returns empty response if query matches any of regular expressions
# target-blacklist = []

[clickhouse]

# see https://clickhouse.tech/docs/en/interfaces/http
url = "http://localhost:8123"

# total timeout to fetch data
data-timeout = "1m0s"
index-table = "graphite_index"
index-use-daily = true
index-reverse-depth = 1

# total timeout to fetch series list from index
index-timeout = "1m0s"

# 'tagged' table from carbon-clickhouse, required for seriesByTag
tagged-table = "graphite_tagged"

# or how long the daemon will query tags during autocomplete
tagged-autocomplete-days = 7

# old index table, DEPRECATED, see description in doc/config.md
# tree-table = ""
# reverse-tree-table = ""
# date-tree-table = ""
# date-tree-table-version = 0
# tree-timeout = "0s"

# is not recommended to use, https://github.com/lomik/graphite-clickhouse/wiki/TagsRU
# tag-table = ""

# add extra prefix (directory in graphite) for all metrics, w/o trailing dot
extra-prefix = ""

# TCP connection timeout
connect-timeout = "1s"

# will be removed in 0.14
# data-table = ""
# rollup-conf = "auto"

# max points per metric when internal-aggregation=true
max-data-points = 1048576

# ClickHouse-site aggregation, see doc/aggregation.md
internal-aggregation = true
```

And I'd like to have it 

```
[common]
# general listener
listen = ":9090"
max-cpu = 1
# limit number of results from find query, 0=unlimited
[...]

PR #541, opened 2021-05-11
Add Encoder.CompactComments to omit extra new line

Felixoid:
**Issue:** #540 

The extra newline before each commit is omitted when `Encoder.CompactComments(true)` is used.

I am not 100% sure about the docstring, so very open to feedback.


pelletier:
Looks good to me, thank you!

Issue #566, opened 2021-07-13
v2 miss Local Time parse 


kkHAIKE:
**Describe the bug**
v2 not support Local Time decode

**To Reproduce**
```
var m map[string]interface{}
err := toml.Unmarshal([]byte(`
[a]
a = 07:32:00
`), &m)
fmt.Println(m, err)
```

**Expected behavior**
get correct result

**Versions**
 - go-toml: 618f018
 - go: 1.16.5
 - operating system: macOS

**Additional context**
```
date-time      = offset-date-time / local-date-time / local-date / local-time
partial-time   = time-hour ":" time-minute ":" time-second [ time-secfrac ]
local-time = partial-time
```
https://github.com/pelletier/go-toml/blob/f6b38c33b7fe522e9c5539f66bd8966ce0006ac6/parser.go#L894-L902

pelletier:
@kkHAIKE since https://github.com/pelletier/go-toml/pull/567 has been merged, is this issue fully resolved for you?

PR #567, opened 2021-07-13
Add LocalTime to interface{} decode support

kkHAIKE:
**Issue:** https://github.com/pelletier/go-toml/issues/566

add localtime decode support


pelletier (changes requested):
Good catch! Thank you for fixing this. I'd love to see some tests before merging this.

pelletier:
Changed the tests to use `Unmarshal` so that it also covers the decoding code.

Thank you for the patch!

PR #568, opened 2021-07-13
hexToString  optimization

kkHAIKE:
simple optimize hexToString..
hexToRune is copy from encoding/json

benchmark
```
func Benchmark_hexToString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		hexToString([]byte("0001F680"), 8)
	}
}
```

```
goos: darwin
goarch: arm64
pkg: tstgo
Benchmark_hexToString
Benchmark_hexToString-8    	30278191	        39.64 ns/op	      16 B/op	       2 allocs/op
Benchmark_hexToString2
Benchmark_hexToString2-8   	134665851	         8.828 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -592,20 +592,19 @@ func (p *parser) parseMultilineBasicString(b []byte) ([]byte, []byte, []byte, er
 			case 't':
 				builder.WriteByte('\t')
 			case 'u':
-				x, err := hexToString(atmost(token[i+1:], 4), 4)
+				x, err := hexToRune(atmost(token[i+1:], 4), 4)
 				if err != nil {
 					return nil, nil, nil, err
 				}
-
-				builder.WriteString(x)
+				builder.WriteRune(x)
 				i += 4
 			case 'U':
-				x, err := hexToString(atmost(token[i+1:], 8), 8)
+				x, err := hexToRune(atmost(token[i+1:], 8), 8)
 				if err != nil {
 					return nil, nil, nil, err
 				}
 
-				builder.WriteString(x)
+				builder.WriteRune(x)
 				i += 8
 			default:
 				return nil, nil, nil, newDecodeError(token[i:i+1], "invalid escaped character %#U", c)
@@ -741,20 +740,20 @@ func (p *parser) parseBasicString(b []byte) ([]byte, []byte, []byte, error) {
 			case 't':
 				builder.WriteByte('\t')
 			case 'u':
-				x, err := hexToString(token[i+1:len(token)-1], 4)
+				x, err := hexToRune(token[i+1:len(token)-1], 4)
 				if err != nil {
 					return nil, nil, nil, err
 				}
 
-				builder.WriteString(x)
+				builder.WriteRune(x)
 				i += 4
 			case 'U':
-				x, err := hexToString(token[i+1:len(token)-1], 8)
+				x, err := hexToRune(token[i+1:len(token)-1], 8)
 				if err != nil {
 					return nil, nil, nil, err
 				}
 
-				builder.WriteString(x)
+				builder.WriteRune(x)
 				i += 8
 			default:
 				return nil, nil, nil, newDecodeError(token[i:i+1], "invalid escaped character %#U", c)
@@ -767,9 +766,14 @@ func (p *parser) parseBasicString(b []byte) ([]byte, []byte, []byte, error) {
 	return token, builder.Bytes(), rest, nil
 }
 
-func hexToRune(s []byte) rune {
+func hexToRune(b []byte, length int) (rune, error) {
+	if len(b) < length {
+		return -1, newDecodeError(b, "unicode point needs %d character, not %d", length, len(b))
+	}
+	b = b[:length]
+
 	var r rune
-	for _, c := range s {
+	for i, c := range b {
 		switch {
 		case '0' <= c && c <= '9':
 			c = c - '0'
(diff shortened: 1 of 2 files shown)


parser.go
```diff
@@ -768,20 +767,31 @@ func (p *parser) parseBasicString(b []byte) ([]byte, []byte, []byte, error) {
...
 
-	//nolint:godox
-	// TODO: slow
-	intcode, err := strconv.ParseInt(string(b), 16, 32)
-	if err != nil {
-		return "", newDecodeError(b, "couldn't parse hexadecimal number: %w", err)
-	}
-
-	return string(rune(intcode)), nil
+	return string(hexToRune(b)), nil
```
pelletier, line 794: When decoding a non-hex character, this function will return `"�", nil`. While before it would return an error. Looks like this should have been caught by a unit test that we should add.

I haven't checked, but if all inputs to this function are already checked to be valid hex characters, the error should be turned into a panic to assert the programming assumption.

pelletier (changes requested):
Thank you for your help on go-toml!

Out of curiosity, do you have benchmarks for this function? The existing benchmarks show no time improvement, but I suspect parsing hex strings represents a minority of the time in those.

Before merging this I'd love to see a better-or-equal execution time and allocations benchmark (I'm fine with equal, just want to check for regressions), as well as handling non-hex characters correctly.

Thank you!

kkHAIKE:
sorry.. hexToString2 test is new function, 8.828 ns/op, 4.5x faster than old one...

pelletier:
Doh! I'm sorry, I somehow didn't see the benchmark in the PR description.

Pushed a commit to benchmark using `parser.parseBasicString` instead of `hexToString`, as it should be closer to what the user experiences. This made me realize that this function was only used in basic (multiline) string parsing, with its result always appended to a string buffer. As a result I changed the signature to return the rune directly, which removes some allocation. Then I inline `hexToRune` to return an error directly from the loop to scope the returned error to the specific invalid character, which should help the user see where the problem is.

---

Benchmarks:

### Baseline vs new hex2string implementation

```
name                           old time/op    new time/op     delta
ParseBasicStringWithUnicode/4     413ns ± 0%      259ns ± 0%  -37.37%  (p=0.000 n=9+9)
ParseBasicStringWithUnicode/8     353ns ± 0%      246ns ± 0%  -30.35%  (p=0.000 n=10+10)

name                           old speed      new speed       delta
ParseBasicStringWithUnicode/4  91.9MB/s ± 0%  146.8MB/s ± 0%  +59.65%  (p=0.000 n=9+9)
ParseBasicStringWithUnicode/8   119MB/s ± 0%    171MB/s ± 0%  +43.57%  (p=0.000 n=10+10)

name                           old alloc/op   new alloc/op    delta
ParseBasicStringWithUnicode/4      112B ± 0%        88B ± 0%  -21.43%  (p=0.000 n=10+10)
ParseBasicStringWithUnicode/8     96.0B ± 0%      80.0B ± 0%  -16.67%  (p=0.000 n=10+10)

name                           old allocs/op  new allocs/op   delta
ParseBasicStringWithUnicode/4      13.0 ± 0%        7.0 ± 0%  -46.15%  (p=0.000 n=10+10)
ParseBasicStringWithUnicode/8      9.00 ± 0%       5.00 ± 0%  -44.44%  (p=0.000 n=10+10)
```

### Baseline vs direct hex2rune call

```
name                           old time/op    new time/op     delta
ParseBasicStringWithUnicode/4     413ns ± 0%      146ns ± 0%   -64.62%  (p=0.000 n=9+10)
ParseBasicStringWithUnicode/8     353ns ± 0%      172ns ± 0%   -51.35%  (p=0.000 n=10+9)

name                           old speed      new speed       delta
ParseBasicStringWithUnicode/4  91.9MB/s ± 0%  259.8MB/s ± 0%  +182.67%  (p=0.000 n=9+10)
ParseBasicStringWithUnicode/8   119MB/s ± 0%    244MB/s ± 0%  +105.52%  (p=0.000 n=10+9)

name                           old alloc/op   new alloc/op    delta
ParseBasicStringWithUnicode/4      112B ± 0%        64B ± 0%   -42.86%  (p=0.000 n=10+10)
ParseBasicStringWithUnicode/8     96.0B ± 0%      64.0B ± 0%   -33.33%  (p=0.000 n=10+10)

name                           old allocs/op  new allocs/op   delta
ParseBasicStringWithUnicode/4      13.0 ± 0%        1.0 ± 0%   -92.31%  (p=0.000 n=10+10)
ParseBasicStringWithUnicode/8      9.00 ± 0%       1.00 ± 0%   -88.89%  (p=0.000 n=10+10)
```

### Side by side

```
name \ time/op                 baseline.txt   branch.txt      branch-2.txt
ParseBasicStringWithUnicode/4     413ns ± 0%      259ns ± 0%      146ns ± 0%
ParseBasicStringWithUnicode/8     353ns ± 0%      246ns ± 0%      172ns ± 0%

name \ speed                   baseline.txt   branch.txt      branch-2.txt
ParseBasicStringWithUnicode/4  91.9MB/s ± 0%  146.8MB/s ± 0%  259.8MB/s ± 0%
ParseBasicStringWithUnicode/8   119MB/s ± 0%    171MB/s ± 0%    244MB/s ± 0%

name \ alloc/op                baseline.txt   branch.txt      branch-2.txt
ParseBasicStringWithUnicode/4      112B ± 0%        88B ± 0%        64B ± 0%
ParseBasicStringWithUnicode/8     96.0B ± 0%      80.0B ± 0%      64.0B ± 0%

name \ allocs/op               baseline.txt   branch.txt      branch-2.txt
ParseBasicStringWithUnicode/4      13.0 ± 0%        7.0 ± 0%        1.0 ± 0%
ParseBasicStringWithUnicode/8      9.00 ± 0%       5.00 ± 0%       1.00 ± 0%
```

pelletier:
I'm going to merge this as is given the significant improvement. Thank you so much for your patch and your help in improving go-toml!

PR #592, opened 2021-09-09
add conf loader

valord577:
Maintain a map for decoded TOML

 - Provide a complete test for https://toml.io
 - Provide a configuration reader like v1

### DIFF
--- a/conf_loader.go
+++ b/conf_loader.go
@@ -1,0 +1,402 @@
+package toml
+
+import (
+	"io"
+	"os"
+	"strings"
+	"time"
+)
+
+var cachedCnf map[string]interface{}
+
+func ReadInString(s string) error {
+	return ReadInBytes([]byte(s))
+}
+
+func ReadInFile(filepath string) error {
+	file, err := os.OpenFile(filepath, os.O_RDONLY, 0444)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = file.Close()
+	}()
+
+	bs, err := io.ReadAll(file)
+	if err != nil {
+		return err
+	}
+
+	return ReadInBytes(bs)
+}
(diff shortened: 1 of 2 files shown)


pelletier:
Thank you for your interest in go-toml! Do you mind expanding on what this code aims to achieve, and the reasons why it should be included in the library? Thanks!

valord577:
There are serveral reasons:

1. Inspired by [Viper](https://github.com/spf13/viper), I am very interested in the way viper manages the configuration. But I expect less dependence. Less is more.
2. I found similar API in v1. As you said, this feature adds complexity in v1. I feel a pity to lose this feature.
3. This is the most important. The code aims to achieve will **NOT** affect the core api. 
4. If these functions are not exported, they can also be used  for testing the toml format.

pelletier:
Thanks for the context! go-toml v2 will be getting a [Document feature](https://github.com/pelletier/go-toml/discussions/471) similar to v1, but significantly more powerful (I'm working on this at the moment).

The proposed change is essentially a set of functions to access data contained in a tree `map[string]interface{}`. Though I understand the usefulness of those functions, I don't think they belong in go-toml. The library already provides a way to go from TOML to `map[string]interface{}` (`Unmarshal`), then the access pattern can be entirely built by the application. Arguably this could even be a dedicated library, as many other systems will generate trees of `map[string]interface{}` that can be used with this set of functions. In some way that's a part of what Viper does.

I agree that less is more. In that case that means keeping go-toml's API lean (this is an explicit design decision for v2).

Not sure I understand the point about testing the toml format.

valord577:
:p

Thanks for your explanation! I made the goal of `go-toml` clear. I will keep interest in `go-toml`!

PR #609, opened 2021-09-27
testsuite: add package for toml-test integration

moorereason:
Does not generate local test files but uses tests embedded in `toml-test` package.  Kept out of internal packages to make `testsuite.Decode` accessible from a future cmd package.

**Discussion:** #599

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,5 +2,8 @@ module github.com/pelletier/go-toml/v2
 
 go 1.16
 
-// latest (v1.7.0) doesn't have the fix for time.Time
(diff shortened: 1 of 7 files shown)


moorereason:
CI runs all tests, so CI is expected to fail.  Failing tests can be skipped by the `tomltest.Runner` in `testsuite/testsuite_test.go`.  Let me know if you want to skip all failing tests for now.

Also, most of this code was adapted from BurntSushi/toml (MIT licensed).  Not sure how you want to handle that.

PR #610, opened 2021-10-01
tomltestgen: add toml-test unit test generation command

moorereason:
* Ported `tomltestgen` to v2.
* `toml-test` file layout changed recently, so updated to work with sub-folders.
* Added go-generate support to the top of `toml_testgen_support_test.go`.
* Tests hidden behind a build tag for now.

**Issue:** #599


pelletier:
Looking good! Do you have plans to import `/testsuite` outside of the module? Otherwise seems like it could go to `/internal` to avoid an accidental import.

Good idea to make it work with `go generate`!

moorereason:
Yes, I have a cmd package for the toml-test binary interface which will use this package.

We could move the generated test file (and support file) into the testsuite package and avoid exporting some functions, but I was trying to work with the existing files.

pelletier:
👍🏻 Sounds good. Merging this and will create a follow up issue to fix failing tests.

Issue #571, opened 2021-07-21
TOML Integer cannot be unmarshalled as a float32


jkongie:
**Describe the bug**
Marshalling a struct with a `Float` value as an integer returns output which when Unmarshalled back into the struct, returns an error `toml: cannot store TOML integer into a Go float32`.

**To Reproduce**
```
package main
import (
	"fmt"
	"github.com/pelletier/go-toml/v2"
)
type Foo struct {
	Float float32
}
func main() {
	foo := Foo{
		Float: 1,
	}
	b, err := toml.Marshal(foo)
	if err != nil {
		fmt.Println(err)
	}
	fmt.Println(string(b))
	// Output: Float = 1
	foo2 := Foo{}
	err = toml.Unmarshal(b, &foo2)
	if err != nil {
		fmt.Println(err)
		// Output: toml: cannot store TOML integer into a Go float32
	}
	fmt.Println(foo2)
}
```

Output
```
Float = 1

toml: cannot store TOML integer into a Go float32
{0}
```

ExpectedOutput
```
Float = 1

{1}
```

**Expected behavior**

It should be able to marshal an struct with a integer as a float, and unmarshal it back into the struct. 

**Versions**
 - go-toml: v2.0.0-beta.3
 - go: version 1.16
 - operating system: macOS

**Additional context**
Add any other context about the problem here that you think may help to diagnose.

pelletier:
Thanks for the report! I wonder if some other conversions are missing.

PR #615, opened 2021-10-04
Fix error with Integer to Float conversion in v2

zostay:
**Issue:** #571 

This is a fix to the integer to float conversion mentioned in #571.

This is breaking a test that seems to be ensuring that the existing behavior is preserved, so it's not ready for a merge. I didn't really know what to make of what this was testing, so if you'll let me know what this test is supposed to accomplish or how to correct it I will.

```
--- FAIL: TestUnmarshalCheckConversionFloatInt (0.00s)
    --- FAIL: TestUnmarshalCheckConversionFloatInt/float (0.00s)
        unmarshal_imported_test.go:1078: 
                Error Trace:    unmarshal_imported_test.go:1078
                Error:          An error is expected but got nil.
                Test:           TestUnmarshalCheckConversionFloatInt/float
FAIL
FAIL    github.com/pelletier/go-toml/v2/internal/imported_tests 0.308s
```

Also, I looked into adding support for unmarshalling into `complex64` and `complex128`, but since you can't marshal those, that seemed a bit pointless. As far as I can tell, this is the only integer conversion missing. I suspect the reverse conversion is not a problem, but I did not look into it.


zostay:
Ah, I misunderstood the problem. Easy fix, though.

pelletier:
Sorry for the late reply! Merged, thank you for the patch!

Issue #613, opened 2021-10-04
Fix tomltestgen failing tests


pelletier:
After merging https://github.com/pelletier/go-toml/pull/610 the toml-test generated tests have been disabled from automatic build because some tests are failing.

This issue tracks fixing the failing tests and ultimately make the toml-test generated tests part of the regular build again.l

jidicula:
I'm taking a crack at `go test -tags testsuite -run TestTOMLTest_Invalid_Array_MissingSeparator` (first entry in the list)

idodin:
Trying out all the comment tests: `go test -tags testsuite -run TestTOMLTest_Invalid_Control_CommentDel`, `go test -tags testsuite -run TestTOMLTest_Invalid_Control_CommentLf`, `go test -tags testsuite -run TestTOMLTest_Invalid_Control_CommentNull` and `go test -tags testsuite -run TestTOMLTest_Invalid_Control_CommentUs`

jidicula:
@idodin I have a WIP for the `TestTOMLTest_Invalid_Control_CommentDel` that I'm about to push up actually. I'll be changing the return of `scanComment()` to return an error if an invalid char is encountered in a comment, so the other comment tests might need that functionality too.

idodin:
@jidicula - sounds good. Do you want to handle the rest of the comments, in that case? I can take a whack at the floats.

jidicula:
Sure thing

stoovon:
Going to take a look at  
```go test -tags testsuite -run TestTOMLTest_Invalid_Table_Injection1```
```go test -tags testsuite -run TestTOMLTest_Invalid_Table_Injection2```
```go test -tags testsuite -run TestTOMLTest_Valid_Table_Keyword```

jidicula:
Working on `go test -tags testsuite -run TestTOMLTest_Valid_Datetime_Local`

moorereason:
I opened a discussion in https://github.com/BurntSushi/toml-test/issues/91 about `valid/datetime/milliseconds` and nanosecond precision.

PR #616, opened 2021-10-05
fix(array test): Fail with missing separator

jidicula:
**Issue:** #613 

The toml-test generated test for an array with missing separators was
failing since the parser wasn't noticing missing separators.

This change leverages the zero value of `ast.Reference` (an int). The
var is now declared outside the while block, and re-zeroed after
checks for array closure and separation have passed.

Prior to this rezeroing, the parser checks if the previous value of
`valueRef` is zero. The only time this check would pass is when the
previous byte in the array was a whitespace, which would only be the
case when an array is missing a separator.


pelletier:
Sorry for the late reply! Thank you for the patch! Adjusted it to return directly when the failure is detected.

PR #618, opened 2021-10-05
Fix failing tests for invalid control characters in comments

jidicula:
**Issue:** #613 

* Modifies `scanComment()` to error out early if an illegal character is found
* Adds illegal characters to `scanComment()`

~~Note that the NUL and US control characters weren't causing tests to fail prior to me making my changes... I took a quick look into why but couldn't crack why those tests weren't failing on my machine (`TestTOMLTest_Invalid_Control_CommentNull` and `TestTOMLTest_Invalid_Control_CommentUs`).~~ Turns out the codegen used the xNN notation instead of Unicode values for NUL and US control characters. I've modified the testgen file accordingly.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -106,7 +106,10 @@ func (p *parser) parseExpression(b []byte) (ast.Reference, []byte, error) {
 	}
 
 	if b[0] == '#' {
-		_, rest := scanComment(b)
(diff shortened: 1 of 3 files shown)


moorereason:
I've offered up a competing PR in #620 that would supersede this offering.

PR #622, opened 2021-10-08
Decode: validate dates

jidicula:
**Issue:** #613 

Constructs a datestring with RFC3339 formatting using the parsed date
values, then attempts to call `time.Parse()` on it, which returns a
non-nil error if datetime string was invalid or `time`'s parsing of
the string failed.

Fixes `go test -tags testsuite -run TestTOMLTest_Invalid_Datetime_ImpossibleDate` failure.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -141,10 +141,6 @@ func parseLocalDateTime(b []byte) (LocalDateTime, []byte, error) {
 	}
 	dt.LocalTime = t
 
-	if !isValidDate(dt.Year, dt.Month, dt.Day) {
-		return LocalDateTime{}, rest, newDecodeError(b, "impossible date")
-	}
-
 	return dt, rest, nil
 }
 


jidicula:
I have some local changes adding similar calls to `time.Parse()` for `parseLocalDate()`, `parseLocalTime()`, and `parseLocalDateTime()` but I'm struggling a bit on where to add tests that adequately cover these changes – should I just be adding cases in `TestUnmarshalDecodeErrors()`?

jidicula:
> Sorry for the late reply! `TestUnmarshalDecodeErrors` should be a good place to put those tests. What error are you getting instead?

```
--- FAIL: TestUnmarshalDecodeErrors (0.00s)
    --- FAIL: TestUnmarshalDecodeErrors/invalid_local_date-time (0.00s)
        unmarshaler_test.go:2003: 
            	Error Trace:	unmarshaler_test.go:2003
            	Error:      	An error is expected but got nil.
            	Test:       	TestUnmarshalDecodeErrors/invalid_local_date-time
FAIL
exit status 1
FAIL	github.com/pelletier/go-toml/v2	0.016s
```

As I understand it, this error tester is expecting an error to be returned, but one isn't being returned even though I'm feeding it a test case with an invalid date.

jidicula:
> Commenting on the general approach taken: this looks like a very expensive approach. Given we have year/month/day already parsed, can we just have a validation step that does not perform allocations?

I was wondering about that, it seemed to me that validating day counts for each month reinvents the wheel a bit. It's definitely easier for time and month validation, though. What approach do you suggest for validating days – perhaps a map that maps month-number to its day count? Or, as I'm typing this I'm thinking this might be an appropriate use for a `switch` statement.

pelletier:
I have no preference on the approach. However the existing tests provide more precise error messages to the user, which is a good property to have. As long as validation makes it clear to the user what the problem is, I don’t have a preference.

jidicula:
Ready for review now

jidicula:
Resolved in 602925b. re: code coverage, are the testgen cases run in CI too now that you've added the skip statements to the still-breaking cases?

jidicula:
> re: code coverage, are the testgen cases run in CI too now that you've added the skip statements to the still-breaking cases?

nvm, I checked and it is. I would guess that the coverage drop is because of the size of the stuff copied over from `time`. I'll add test cases for those.

jidicula:
Good call with the coverage comment, I looked deeper and found some unreachable code in a redundant date validity check 🙂. Should be good for re-review now @pelletier.

pelletier:
Looking good, merged! Thank you!

jidicula:
> Looking good, merged! Thank you!

My pleasure! If it's not too much trouble, could you put a `hacktoberfest-accepted` label on this PR? They don't seem to be tracking review acceptance and merges correctly.

jidicula:
> No problem, done. I bet they only look at PRs merged in the main branch, which v2 is not quite ready for that yet.

Ah, good point. Thanks!

PR #673, opened 2021-11-23
Decoder: flag invalid carriage returns in literal strings

moorereason:
Add `\r` to disallowed new line control characters in literal strings.


pelletier:
Thank you for fixing this!

PR #696, opened 2021-12-02
Decode: restrict timezone offset values

moorereason:
Don't allow hours greater than 24 and minutes greater than 60 per RFC
3339.


pelletier:
Thank you!

PR #699, opened 2021-12-05
Decode: ensure signed exponents don't start with an underscore

moorereason:


pelletier:
Thanks for the patch! Not found with fuzzing?

moorereason:
Everything I've found recently has been from fuzzing. 😄

PR #700, opened 2021-12-05
Decode: allow maximum seconds value of 60

moorereason:
RFC3339 allows seconds to be 60 when adding leap seconds


pelletier:
Didn't know! Thank you.

PR #706, opened 2021-12-22
Decode: fix maximum time offset values

moorereason:
According to RFC3339 section 5.6, the maximum time offset values for
hours and minutes is 23 and 59, respectively.

Found while doing differential fuzzing against toml-dart.


pelletier:
Thank you!

Issue #719, opened 2021-12-30
Port jsontoml to v2


pelletier:

pelletier:
Fixed in https://github.com/pelletier/go-toml/commit/d8ddc00c61af538336d851a07893a671f8f37174

PR #726, opened 2021-12-31
jsontoml: port to v2

pelletier:
Fixes https://github.com/pelletier/go-toml/issues/719


pelletier:
Coverage went down because I am not nesting the `usage()` and `os.Exit` calls, which is fine.

PR #738, opened 2022-01-21
marshaler: flatt anonymous structs

j-licht:
rebuild behavior of handling anonymous structs like in v1.
Also copied the both tests from v1.
Without this patch unmarshaling of a marshaled struct with anonymous
structs isn't possible.


pelletier:
Hi! Closing this in favor of https://github.com/pelletier/go-toml/pull/747. You are correct that the current behavior of never flattening is wrong, but there is a bit more subtlety to it. Let me know if it works for you!

PR #750, opened 2022-04-08
Decode: allow subtables to be defined later in the document

### LINKED ISSUE #739: Decode: incorrect table redefinition error
> table bar already exists

```text
[foo.bar.baz]
wibble = 'wobble'

[foo]

[foo.bar]
huey = 'dewey'
```

pelletier:
Fixes #739


pelletier:
Coverage went down because the `internal/tracker` finally got a test, but the CI script doesn't add up coverage across packages.

PR #770, opened 2022-05-03
Typo in README.md fix

fnork:
Change "document was not prevent in the target structure" to "document was not present in the target structure".

No code changes so `benchstat` results remain unchanged.


pelletier:
Thank you for catching this!

PR #809, opened 2022-08-22
Add back missing build tag needed for Go 1.16 and earlier

piotrbulinski:
Per https://pkg.go.dev/cmd/go#hdr-Build_constraints:

> Go versions 1.16 and earlier used a different syntax for build constraints, with a "// +build" prefix. The gofmt command will add an equivalent //go:build constraint when encountering the older syntax.

Adding the missing "+build" comment, so builds on go 1.16 work.

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -1,4 +1,5 @@
 //go:build go1.18 || go1.19
+// +build go1.18 go1.19
 
 package toml_test
 


pelletier:
Good catch, thank you!

piotrbulinski:
@pelletier thanks for merging. Do you have a plan for releasing a new version (which would include this change)? Some of my projects builds are currently failing on go 1.16 due to this problem and it would be great to get them running without any workarounds. :)

pelletier:
Done :) https://github.com/pelletier/go-toml/releases/tag/v2.0.4

PR #818, opened 2022-09-09
Fix reflect.*Header in TinyGo.

eNV25:
TinyGo uses uintptr for Len and Cap in reflect.*Header, unlike gc which uses int. This commit chooses the correct type using -tags.

For unrelated reasons, compiling with TinyGo does not yet work. But this eliminates one error.

### DIFF
--- a/internal/danger/danger.go
+++ b/internal/danger/danger.go
@@ -2,11 +2,10 @@ package danger
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"unsafe"
 )
-
-const maxInt = uintptr(int(^uint(0) >> 1))
 
 func SubsliceOffset(data []byte, subslice []byte) int {
 	datap := (*reflect.SliceHeader)(unsafe.Pointer(&data))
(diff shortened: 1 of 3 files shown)


internal/danger/danger.go
```diff
@@ -17,21 +16,19 @@ func SubsliceOffset(data []byte, subslice []byte) int {
 	}
 	offset := hlp.Data - datap.Data
 
-	if offset > maxInt {
+	if offset > math.MaxInt {
```
pelletier: `math.MaxInt` only exists in Go 1.17+. Do you mind moving the current definition of maxInt in the `sizetype.go` file?

Officially go-toml doesn't support versions older than the last two Go versions, but I feel like in this case it would be a small effort to keep the compatibility for older versions.

pelletier:
Thanks for the patch! I'd love for go-toml to support tinygo. I'll try to see if I can set up CI to test against it too.

internal/danger/danger.go
eNV25: > `math.MaxInt` only exists in Go 1.17+. Do you mind moving the current definition of maxInt in the `sizetype.go` file?

Sounds good to define privately. It's probably a better idea to do it in `danger.go` itself, to not define it twice in `sizetype.gp` and `sizetype_tinygo.go`.

PR #837, opened 2023-01-10
Encode: fix inline table first key value whitespace

cuonglm:
When encoding table inline, we should not apply ident. Otherwise, the key-value will be right align.

For example, this following data:

```
x := map[string][]map[string]string{
	"one": []map[string]string{
		{"0": "0"},
		{"1": "1"},
	},
}
```

will be encoded as:

```
one = [
  {  0 = '0'},
  {  1 = '1'}
]
```

This PR makes is to be encoded as:

```
one = [
  {0 = '0'},
  {1 = '1'}
]
```


pelletier:
Looks good, thank you for the patch! Sorry it took me a while to find time to review.

PR #578, opened 2021-09-03
feat(tomll): add multiLineArray flag to linter

rasta-rocket:
Hi :wave:

First of all, thanks for your work on this project :muscle:.

We use the linter in our CI and thanks to that we have been able to keep the quality of our TOML files.

Since things started to get serious, we have moore and more bigger arrays that became hard to maintain since the linter keep the arrays in one line.

### Explanation of what this pull request does.

The goal of this PR is to add a flag to the linter that allow us to render the TOML file with multulines array (meaning one line by element)

Don't hesitate if you need more details

Cheers :wink:

PR #838, opened 2023-01-15
Add multi line array options

rasta-rocket:
Hi @pelletier 👋 

It has been a while since my last PR #578 on this repo.

Well the goal here is to re add that feature in the v2 of `go-toml` and especially `tomll`

### Explanation of what this pull request does.

The goal of this PR is to add a flag to the linter that allow us to render the TOML file with multilines array (meaning one line by element)

Don't hesitate if you need more details

Cheers 😉

---

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -12,7 +12,9 @@ import (
 	"github.com/pelletier/go-toml/v2"
 )
 
-type ConvertFn func(r io.Reader, w io.Writer) error
+type Options map[string]interface{}
+
+type ConvertFn func(r io.Reader, w io.Writer, options Options) error
 
 type Program struct {
 	Usage string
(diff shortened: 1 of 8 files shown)


internal/cli/cli.go
```diff
@@ -12,14 +12,17 @@ import (
 	"github.com/pelletier/go-toml/v2"
 )
 
-type ConvertFn func(r io.Reader, w io.Writer) error
+type Options map[string]interface{}
```
pelletier, line 15: For now I'd like this to be a `struct`, to lean on go's default values to avoid exists/empty checks and casting, and make it easier to understand which options are supported.

Because `Program` is used by multiple programs, a single `Options` struct may be unlikely to work with all of them. I think the easiest way around this is to have `convert()` take an options struct, and return a function, creating a closure around the provided options. That way each program can build its own options an generate its own processing function, without the need for a one-size-fits-all options type.

pelletier (commented):
Thank you for the pull request! and sorry it took me so long to get back to you.

I'd like to see if we can use a closure around each program's processing function to handle arguments locally instead of a generic way of passing options.

Issue #839, opened 2023-01-20
panic unmarshaling array of tables


jmank88:
**Describe the bug**
Unmarshaling to a field of a table in an array of tables causes a panic.

**To Reproduce**
https://go.dev/play/p/l2KDjY4lX-3
```go
m := struct {
	A []struct {
		B struct{}
	}
}{}
fmt.Println(toml.Unmarshal([]byte(`[[A.B]]`), &m))
```

**Expected behavior**
If this is invalid, I expect an error rather than a panic.

**Versions**
 - go-toml: 2.0.6
 - go: 1.19
 - operating system: Linux, Go-playground

pelletier:
Thank you for the bug report! I'll take a look when I have time, but feel free to send a pull request in the meantime!

jmank88:
@pelletier I started a draft that highlights the problematic area (#840). Is returning an error the correct thing to do here?

pelletier:
You're correct: this should be an error, because the array of tables is on `A.B`, and `B` is a struct, not a slice.

pelletier:
Should be fixed now.

PR #840, opened 2023-02-04
panic unmarshaling array of tables

jmank88:
Towards #839
- [x] `TestUnmarshalArrayOfTables` to reproduce
- [x] Placeholder error instead of panic to 'fix' the test
- [ ] Better error? _Or_ is this actually valid, meaning e.g. the slice should be expanded instead?

### DIFF
--- a/unmarshaler.go
+++ b/unmarshaler.go
@@ -377,7 +377,11 @@ func (d *decoder) handleArrayTableCollection(key unstable.Iterator, v reflect.Va
 
 		return v, nil
 	case reflect.Slice:
-		elem := v.Index(v.Len() - 1)
+		idx := v.Len() - 1
+		if idx < 0 {
+			return v, fmt.Errorf("TODO uninitialized array table")
+		}
+		elem := v.Index(idx)
 		x, err := d.handleArrayTable(key, elem)
 		if err != nil || d.skipUntilTable {
 			return reflect.Value{}, err
--- a/unmarshaler_test.go
+++ b/unmarshaler_test.go
@@ -11,9 +11,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/pelletier/go-toml/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/pelletier/go-toml/v2"
 )
 
 func ExampleDecoder_DisallowUnknownFields() {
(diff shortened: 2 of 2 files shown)


unmarshaler.go
```diff
@@ -377,7 +377,11 @@ func (d *decoder) handleArrayTableCollection(key unstable.Iterator, v reflect.Va
 
 		return v, nil
 	case reflect.Slice:
-		elem := v.Index(v.Len() - 1)
+		idx := v.Len() - 1
+		if idx < 0 {
```
pelletier, line 381: This seems odd to me. I'd expect something like a type check, so that we can confidentially return an error along the lines of "you cannot put an array of tables into something that is not a slice or an array" (there may be better wording 😅 )

jmank88: IIUC this index panic must be dealt with first. Perhaps my original example was not minimal enough. Consider this version, with the only change being an explicit `[[A]]` before `[[A.B]]`:
https://go.dev/play/p/rJqxTrNgONH
> toml: cannot decode TOML array table into struct field struct { B struct {} }.B of type struct {}

Should the slice be expanded instead, so that we proceed along to this error ^?

jmank88: Here is a better version:
Panic: https://go.dev/play/p/ERU8JP5EldZ
No Panic with explicit `[[A]]`: https://go.dev/play/p/KL1MKmdEf6k

pelletier: Implicitly, `A` is a table. Adding an explicit `[[A]]` actually changes the structure of the document to make it an array of tables, so I don't think we can compare those.

To me, the logical path of the unmarshaler is:

1. Is `A` a key in the current target? -> ✅ it's a field in the root struct. That value becomes the current target.
2. Is `B` a key in the current target? -> ❎ error: the current target is a slice, so B not a key; and `A` hasn't been marked as an array table.

The problem is, the unmarshaler doesn't explicitly tracks which tables are array tables or not for the sake of performance (except in strict mode). However, when it encounters an explicit array table, it guarantees that the slices have been expended as needed to be ready to receive a key lookup. This is because `[[A]]` is equivalent to `{"A": [{}]}` in json: https://go.dev/play/p/74TSvAtr8_R.

So, we should be able to use the fact that the unmarshaler is finding itself at a point where the container is a slice but hasn't been allocated as an indicator that the type are incompatibles, and document is possibly missing an explicit array table.

pelletier:
Bug was fixed recently, sorry.

PR #856, opened 2023-03-04
Add errorcontext to unmarshalString()

Krishks369:
This PR tries to solve the issue [here](https://gitlab.com/gitlab-org/gitaly/-/issues/4803) 

@pellitier I have added the error context to the function according to my knowledge and please do check if its correct. Also can you provide a way to test it if its correct. Thanks for the reply

### DIFF
--- a/unmarshaler.go
+++ b/unmarshaler.go
@@ -979,11 +979,17 @@ func (d *decoder) unmarshalString(value *unstable.Node, v reflect.Value) error {
 	case reflect.Interface:
 		v.Set(reflect.ValueOf(string(value.Data)))
 	default:
+		ctx := d.errorContext
+		if ctx != nil && ctx.Struct != nil {
+			f := ctx.Struct.FieldByIndex(ctx.Field)
+			return unstable.NewParserErrorWithContext(d.p.Raw(value.Raw), "cannot store TOML string into struct field %s.%s of type %s", f.Name, f.Type, v.Kind())
+		}
 		return unstable.NewParserError(d.p.Raw(value.Raw), "cannot store TOML string into a Go %s", v.Kind())
 	}
 
 	return nil
 }
+
 
 func (d *decoder) handleKeyValue(expr *unstable.Node, v reflect.Value) (reflect.Value, error) {
 	d.strict.EnterKeyValue(expr)


unmarshaler.go
```diff
@@ -979,12 +979,18 @@ func (d *decoder) unmarshalString(value *unstable.Node, v reflect.Value) error {
 	case reflect.Interface:
 		v.Set(reflect.ValueOf(string(value.Data)))
 	default:
+		ctx := d.errorContext
+		if ctx != nil && ctx.Struct != nil {
+			f := ctx.Struct.FieldByIndex(ctx.Field)
+			return unstable.NewParserErrorWithContext(d.p.Raw(value.Raw), "cannot store TOML string into struct field %s.%s of type %s", f.Name, f.Type, v.Kind())
```
pelletier, line 985: Unless I missed something, you should be able to use `d.typeMismatchError()` to generate an error that resulted from a mismatch between the expected Go type and the type found in the TOML document. It will take care of creating the human readable error based on whether the decoder is targeting a struct.

arkn98: Hey @pelletier,

Pardon my ignorance. I'm a newbie. Is there a way to get a human-readable error string when the type of a field mismatches? `d.typeMismatchError()` is internal, and I can't use it in my code. And, `DecodeError.String()` is multi-line, which is not something I want.

This is what I'm trying:
If my config looks like this,
```
[server]
path = "/my/path"
port = 50
```

and, I read it like this,
```
if err := toml.NewDecoder(file).Decode(&cfg); err != nil {
    return Cfg{}, fmt.Errorf("load toml: %w", err)
}
```
I'm getting the error message `load toml: cannot store TOML string into a Go int` when the `port` is set to a string. How can I get the error message to mention the field `server.port` that is causing the problem?

pelletier: @arkn98 sorry I didn't see this. You're doing it right; the error message should contain the name of the field causing problem. Seems like this specific error is not constructed properly to include it.

pelletier:
@Krishks369 @arkn98 This should now be fixed by https://github.com/pelletier/go-toml/pull/891. Feel free to reopen this PR or create an issue if not!

unmarshaler.go
arkn98: No worries at all. Your fix seems to work. Let me test it out. Thanks for looking into this!

PR #861, opened 2023-04-22
fuzz: move fuzz_target from oss-fuzz

manunio:
- This pr attempts to move `fuzz_toml` fuzz_target from oss-fuzz to go-toml.
- Its recommended by oss-fuzz to move fuzz-target upstream as this eases maintenance of fuzz_target.
- It also include check against too large input.
- Skips fuzz from coverage report.
- Adds a new ossfuzz package

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -77,7 +77,7 @@ cover() {
 
     pushd "$dir"
     go test -covermode=atomic  -coverpkg=./... -coverprofile=coverage.out.tmp ./...
-    cat coverage.out.tmp | grep -v testsuite | grep -v tomltestgen | grep -v gotoml-test-decoder > coverage.out
+    cat coverage.out.tmp | grep -v fuzz | grep -v testsuite | grep -v tomltestgen | grep -v gotoml-test-decoder > coverage.out
     go tool cover -func=coverage.out
     popd
 
--- a/oss_fuzz.go
+++ b/oss_fuzz.go
@@ -1,0 +1,43 @@
+//go:build go1.18 || go1.19 || go1.20
+// +build go1.18 go1.19 go1.20
+
+package toml
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+func FuzzToml(data []byte) int {
+	if len(data) >= 10240 {
+		return 0
+	}
+
+	if strings.Contains(string(data), "nan") {
+		return 0
+	}
+
+	var v interface{}
+	err := Unmarshal(data, &v)
+	if err != nil {
+		return 0
+	}
+
+	encoded, err := Marshal(v)
+	if err != nil {
+		return 0
+	}
+
+	var v2 interface{}
+	err = Unmarshal(encoded, &v2)
+	if err != nil {
+		panic(fmt.Sprintf("Failed round trip: %s", err))
+	}
+
+	if !reflect.DeepEqual(v, v2) {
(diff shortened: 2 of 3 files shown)


manunio:
- fuzzing check: It will be fixed at oss-fuzz as soon as this pr lands.

ci.sh
```diff
@@ -77,7 +77,7 @@ cover() {
 
     pushd "$dir"
     go test -covermode=atomic  -coverpkg=./... -coverprofile=coverage.out.tmp ./...
-    cat coverage.out.tmp | grep -v testsuite | grep -v tomltestgen | grep -v gotoml-test-decoder > coverage.out
+    cat coverage.out.tmp | grep -v fuzz | grep -v testsuite | grep -v tomltestgen | grep -v gotoml-test-decoder > coverage.out
```
pelletier, line 80: Any reason for removing fuzz output?

oss_fuzz.go
```diff
@@ -0,0 +1,43 @@
...
+
+	var v interface{}
+	err := Unmarshal(data, &v)
+	if err != nil {
+		return 0
+	}
+
+	encoded, err := Marshal(v)
+	if err != nil {
+		return 0
```
pelletier: Should this be a panic? It seems like a piece of that returned by `Unmarshal` should be able to go through `Marshal` without error.

ci.sh
manunio: Before removing fuzz output, coverage report ci was failing with following error:
```

Results: v2 97.1% HEAD 96.3%
Delta: -.8
Regression!
							      >	FuzzToml	0.0%
```
as a fix i tried moving oss_fuzz.go to oss_fuzz_test.go as coverage seems to ignore _test files but  it showed a different error:
```
wrong signature for FuzzToml, must be: func FuzzToml(f *testing.F)
```

manunio: Do you wish to revert that ?

oss_fuzz.go
manunio: That makes sense, will do that :)

manunio:
Friendly Ping :)

pelletier:
Sorry for the delay!

I think I know what's going on. When the code is in `oss_fuzz.go`, the coverage assumes it's part of the codebase. When it's part of the `_test.go` file, Go expects all `Fuzz*` functions to have a specific signature for the built-in fuzzing mechanism.

I just realized that the current proposal would add `FuzzToml` to the public API of go-toml, which I'd like to avoid. Is there a way to either put it in a new `/ossfuzz` directory (if you need to call the function directly) or `/cmd/ossfuzz` if you'd rather use a binary.

manunio:
> Sorry for the delay!
> 
> I think I know what's going on. When the code is in `oss_fuzz.go`, the coverage assumes it's part of the codebase. When it's part of the `_test.go` file, Go expects all `Fuzz*` functions to have a specific signature for the built-in fuzzing mechanism.
> 
> I just realized that the current proposal would add `FuzzToml` to the public API of go-toml, which I'd like to avoid. Is there a way to either put it in a new `/ossfuzz` directory (if you need to call the function directly) or `/cmd/ossfuzz` if you'd rather use a binary.

Hi, Thanks for the feedback, I have made changes as per your request :)

pelletier (approved):

pelletier:
Thank you!

PR #864, opened 2023-05-12
fuzz: improve target perf

manunio:
- This attempts to improve issue types like timeout, slow_unit and speed, as seen in following oss-fuzz performance report: https://oss-fuzz.com/performance-report/libFuzzer_go-toml_fuzz_toml/libfuzzer_asan_go-toml/2023-05-11


pelletier:
Sounds good. Let's not go down lower than that though. This may be hiding some kind of performance issue or infinite loop bug.

Issue #857, opened 2023-03-10
default values for objects in variable length arrays


cfal:
Hi there, I realize that v2 doesn't support default values - but I'm wondering the best way to do this, or if it's possible even in combination with go-defaults or other custom code.

I am trying to decode a TOML document that contains a variable length array of objects - and I'd like to prefill the field values as recommended in the readme.

eg, in this example, i'd like C to be set to `-1` when the key is not present:

```
package main

import (
        "bytes"
        "fmt"
	"github.com/pelletier/go-toml/v2"
)

func main() {
	type varData struct {
	        A int
	        B int
	        C int
	}
	type data struct {
		VarData []varData
	}
	
	var d data
	tomlBlob := []byte(`
[[VarData]]
A = 1
B = 2

[[VarData]]
A = 2
B = 1
`)

	toml.NewDecoder(bytes.NewReader(tomlBlob)).Decode(&d)
	fmt.Printf("%+v\n", d)
}
```

it seems like with a library like go-defaults, you'd still need to call a `SetDefaults()` method, and there isn't a good way to do this in advance for variable length array objects.

An option seems to be to use `encoding.TextUnmarshaler` perhaps? i was thinking something like this:

```
func (v *VarData) UnmarshalText(text []byte) error {
  v.SetDefaults()
  return toml.NewDecoder(text).Decode(&v)
}
```

..but I was also curious with the TextUnmarshaler approach (if it works?), how you would know that the provided `text []bytes` parameter is a TOML string and due to a go-toml invocation, and not from decoding other formats - ie, this would error out if we were trying to decode a JSON string instead.

Thank you!

cfal:
i stumbled upon #484 and [its test](https://github.com/pelletier/go-toml/blob/643c251c4b34637fef522646c3513598cd4b90f7/unmarshaler_test.go#L2052), and noticed that in fact `UnmarshalText` does not get called when items are defined in this manner:

```
package main

import (
	"fmt"
	"github.com/pelletier/go-toml/v2"
	"strconv"
)

type Integer struct {
	Value int
}

func (i Integer) MarshalText() ([]byte, error) {
	return []byte(strconv.Itoa(i.Value)), nil
}
func (i *Integer) UnmarshalText(data []byte) error {
	fmt.Println("NEVER CALLED")
	conv, err := strconv.Atoi(string(data))
	if err != nil {
		return err
	}
	i.Value = conv
	return nil
}

type Config struct {
	Integers []Integer
}

func main() {
[...]

PR #870, opened 2023-05-19
Fix UnmarshalTOML is not called when its underlying type is an array

ImSingee:
This PR is for v1.

This resolves the issue where, if a type implements `toml.Unmarshaler` but it has an array as its underlying type, the `UnmarshalTOML` method won't be triggered.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -1017,6 +1017,15 @@ func (d *Decoder) valueFromToml(mtype reflect.Type, tval interface{}, mval1 *ref
 		return reflect.ValueOf(nil), fmt.Errorf("Can't convert %v(%T) to trees", tval, tval)
 	case []interface{}:
 		d.visitor.visit()
+
+		// Check if pointer to value implements the Unmarshaler interface.
+		if mvalPtr := reflect.New(mtype); isCustomUnmarshaler(mvalPtr.Type()) {
+			if err := callCustomUnmarshaler(mvalPtr, tval); err != nil {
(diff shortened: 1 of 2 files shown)


pelletier:
Hi! Thank you for the patch! However v1 is not supported anymore and will not receive any update. Consider using go-toml v2 instead.

ImSingee:
@pelletier Thanks for yor reminder. You can close this PR at your opinion.

I’m not using v2 (in fact, I just downgraded to v1 today) because and only because of the lack of missing toml.Unmarshaler interface.

Though there’s TextUnmarshaler, but it can only handle string fields. However, I want to (and need to) do more for historical reasons. The case is just that we have a field that needs to accept both a string and a string array. But in v2, there’s no way to implement it.

Is it possible to re-add the toml.Unmarshaler interface? It will do increase flexibility. And if you accept the feature, I can submit a PR to implement that.

pelletier:
Makes sense. I'd be happy to see a good patch to implement `toml.Unmarshaler` for v2! I think it would also help others (https://github.com/pelletier/go-toml/issues/857).

ImSingee:
@pelletier 👀 Okay, I will try to implement it!

pelletier:
@ImSingee created https://github.com/pelletier/go-toml/issues/873 with some thoughts on the topic, hopefully it's helpful!

pelletier:
Should be fixed now, sorry.

PR #876, opened 2023-05-29
Encode: fix support for arrays

MrJetBOX:
This change allows encoding fixed size arrays. 
Current version generates error: `toml: cannot encode value of type array`.


pelletier:
Can't believe I missed that! Thanks a lot for the patch!

PR #879, opened 2023-06-03
Decode: assign empty struct to empty defined sections

### LINKED ISSUE #878: omitempty returns nil for defined empty sections
**Describe the bug**

When parsing a document into a struct using pointers and `omitempty`, empty sections are interpreted as `nil` same as if they weren't defined at all.

The main issue with this behavior is that the action `Marshal(Unmarshal(doc))` mutates the document.

**To Reproduce**

Run the following example:

```go
package main

import (
	"fmt"

)

type A struct {
	Nullable *struct {
		A string
	} `toml:"nullable,omitempty"`
}

type Doc struct {
	A *A `toml:"a,omitempty"`
}

func main() {
	doc1 := []byte(`[a]
[a.nullable]
`)
	doc2 := []byte(`[a]`)

	var d1 Doc
	if err := toml.Unmarshal(doc1, &d1); err != nil {
		panic(err)
	}
	fmt.Println(d1.A.Nullable)

	var d2 Doc
	if err := toml.Unmarshal(doc2, &d2); err != nil {
		panic(err)
	}
	fmt.Println(d2.A)
}
```

which prints:

```
<nil>
<nil>
```

**Expected behavior**

I'd expect it to return `&Nullable{}` in the first case as the section was defined.

**Versions**
 - go-toml: 60e4af8
 - go: 1.20
 - operating system: Linux

**Additional context**

dbarrosop:
Fixes #878 

Not sure what the gate in the patched method was intended to do but it was preventing from creating pointers to structs correctly when a section was defined. This change ensures that `original_document == Marshal(Unmarsal(original_document)` which isn't the case without this patch.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -79,6 +79,7 @@ cover() {
     go test -covermode=atomic  -coverpkg=./... -coverprofile=coverage.out.tmp ./...
     cat coverage.out.tmp | grep -v fuzz | grep -v testsuite | grep -v tomltestgen | grep -v gotoml-test-decoder > coverage.out
     go tool cover -func=coverage.out
+    echo "Coverage profile for ${branch}: ${dir}/coverage.out" >&2
     popd
 
     if [ "${branch}" != "HEAD" ]; then
--- a/unmarshaler.go
+++ b/unmarshaler.go
@@ -1170,10 +1170,10 @@ func initAndDereferencePointer(v reflect.Value) reflect.Value {
 
 // Same as reflect.Value.FieldByIndex, but creates pointers if needed.
 func fieldByIndex(v reflect.Value, path []int) reflect.Value {
-	for i, x := range path {
+	for _, x := range path {
 		v = v.Field(x)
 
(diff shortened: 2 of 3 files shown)


pelletier:
Thanks a lot for the patch, and sorry for the delay! Not sure what that check was about either, but I think having round-trip equality is a good property to have.

pelletier:
Looks like removing this check either made some code unused, or created untested cases. Do you mind taking a look and removing the code or adding relevant tests so that we are sure we don't miss anything? Otherwise, I can check next weekend.

dbarrosop:
Sorry for the delay, I was travelling during the week. I tried looking into this but I am afraid I am not familiar enough with the codebase to understand why the coverage went down or how to increase it back. If this is something you could help with I'd really appreciate it. Thanks!

dbarrosop:
@pelletier do you think you could help with the coverage, please?

pelletier:
Done; thank you for the nudge!

dbarrosop:
Awesome! Thanks for the help :)

Issue #882, opened 2023-07-17
Ordered encoded results is incorrect in some cases


msaf1980:
**Describe the bug**
Ordered encoded results is incorrect in some cases (simple type after map or complex slice)

**To Reproduce**

Test for reproduce

```
func TestMarshalOrder(t *testing.T) {
	type included struct {
		Inc int `comment:"Included"`
	}
	type order struct {
		One   int        `comment:"Before array"`
		Two   []included `comment:"Array"`
		Three map[string]time.Duration
		Four  bool `comment:"Order check"`
		Five  []int
	}

	doc := order{
		One: 1, Two: []included{{4}, {2}}, Three: map[string]time.Duration{"3": time.Second}, Four: true, Five: []int{5, 2},
	}

	expected := `
# Before array
One = 1

# Order check
Four = true
Five = [5, 2]

[[Two]]

 # Included
 Inc = 4

[[Two]]

 # Included
 Inc = 2

[Three]
 3 = "1s"
`

	var resultDoc order

	result := new(bytes.Buffer)
	encoder := NewEncoder(result).Indentation(" ").Order(OrderPreserve)
	err := encoder.Encode(doc)
	require.NoError(t, err)

	err = Unmarshal(result.Bytes(), &resultDoc)
	assert.NoError(t, err)

	require.Equal(t, expected, result.String())
	require.Equal(t, doc, resultDoc)
}
```

**Expected behavior**
Encode simply type first, then complex.

**Versions**
 - go-toml: master (5dc006f)
 - go: All
 - operating system: All

pelletier:
Is there something blocking you from upgrading to go-toml v2? This bug has been fixed in that version ([playground](https://go.dev/play/p/Cjm-9fVnYtD)), and go-toml v1 is not receiving updates anymore.

msaf1980:
> Is there something blocking you from upgrading to go-toml v2? This bug has been fixed in that version ([playground](https://go.dev/play/p/Cjm-9fVnYtD)), and go-toml v1 is not receiving updates anymore.

v2 not supporting tag `commented`. It's main reason why we can't migrate to v2.

pelletier:
Can you elaborate on your use-case of the `commented` tag? I'm curious how it's used to understand whether that problem would be better solved by the document model, of if it's worth re-introducing the `commented` flag.

msaf1980:
It's used for generate commneted config files in our projetcs.

pelletier:
Could you provide an example?

[...]

PR #883, opened 2023-07-17
Fix ordered encoding

msaf1980:
Fix for issue #882 (Ordered encoded results is incorrect in some cases (simple type after map or complex slice)


pelletier:
Closing this as go-toml v1 isn't receiving any update, and [2.1.0 has been released](https://github.com/pelletier/go-toml/discussions/895), which brings back `commented` to allow you to update to it.

Issue #886, opened 2023-08-21
Panic when parsing '0' using parseFloat


ocean2811:
**Describe the bug**
Panic when parsing '0'(number without decimals) as a float type.

**To Reproduce**
```go
const file = `foo = 0` // read from .toml file

type Toml struct {
    Foo float64 `toml:"foo"`
}

t := &Toml{}
err := toml.Unmarshal([]byte(file), t)
fmt.Println(t, err)
return
```

**Expected behavior**
Normal output of `TOML` structure contents.

**Versions**
 - go-toml: v2.0.9 (e183db7e69f2c8da8ddff8be040ce1f79024d075)
 - go: version 1.20.4
 - operating system: Windows, Linux

PR #887, opened 2023-08-21
Decode: fix panic when parsing '0' as a float

ocean2811:
Fix (#886) :  Panic when parsing '0' using parseFloat


pelletier:
Surprising that this bug has been around for so long! Thank you for the report and the patch.

PR #892, opened 2023-08-28
Decode: fix wrong indention for comments on tables

Felixoid:
Follow up for #889, now comments for tables have an indentation of the parent


pelletier:
Looks good, thank you for the fix!

PR #907, opened 2023-10-13
Add cmd/gotoml-test-encoder

arp242:
Since there was already an encoder I figured adding an encoder makes sense. Also had to fix it for local dates.

	% toml-test -encoder ./gotoml-test-encoder
	toml-test v2023-10-13 [./gotoml-test-encoder]: using embedded tests
	encoder tests: 162 passed,  0 failed

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -77,7 +77,7 @@ cover() {
 
     pushd "$dir"
     go test -covermode=atomic  -coverpkg=./... -coverprofile=coverage.out.tmp ./...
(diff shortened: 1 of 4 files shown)


arp242:
I copied all the extra tests from testdata/fuzz/FuzzUnmarshal to toml-test by the way:
https://github.com/toml-lang/toml-test/commit/f03077ad42257217c30fb7969ee141a892f04a3b

Updating toml-test fails though, since it will also include the TOML 1.1 tests; you'll need to either use the `tests/files-toml-1.0.0` file, or use `toml-test -copy` from the CLI.

pelletier:
Thank you for taking care of this! I've created a follow-up issue to update toml-test properly.

PR #911, opened 2023-10-25
use toml-test to generate tests

mpldr:
This patch enables using toml-test to generate the testfiles. The tests currently generated fail to compile, but as far as I can tell at 12:40 am this seems not to be caused by me. I at least hope it's something trivial to fix.


pelletier:
Thank you for taking a stab at this!

> The tests currently generated fail to compile

Tests don't compile because some of them are generated multiple times in `toml_testgen_test.go`. For example:

```
$ ag TestTOMLTest_Invalid_NoClose1
toml_testgen_test.go
38:func TestTOMLTest_Invalid_NoClose1(t *testing.T) {
733:func TestTOMLTest_Invalid_NoClose1(t *testing.T) {
1578:func TestTOMLTest_Invalid_NoClose1(t *testing.T) {
1708:func TestTOMLTest_Invalid_NoClose1(t *testing.T) {
```

I haven't looked into it but seems like the new way to generate test name isn't unique enough?

mpldr:
On Thu Oct 26, 2023 at 7:19 PM CEST, Thomas Pelletier wrote:
> I haven't looked into it but seems like the new way to generate test name isn't unique enough?

That was also my guess, but I haven't taken a look at the name
generation at all.

-- 
Moritz Poldrack
https://moritz.sh

> Made with real ingredients.

go.mod
```diff
@@ -2,4 +2,7 @@ module github.com/pelletier/go-toml/v2
 
 go 1.16
 
-require github.com/stretchr/testify v1.8.4
+require (
+	github.com/stretchr/testify v1.8.4
+	github.com/toml-lang/toml-test v1.4.1-0.20231018175634-23c30e9eb082
```
pelletier: I would really like to avoid adding dependencies to go-toml. That's the reason why zip files were used in the first place. I think two options are either still download zips and filter out tests for TOML > 1.0, or create a submodule for the test generator with its own `go.mod`, to avoid imposing those dependencies on users of go-toml.

mpldr: In that case, I would probably suggest doing the latter. I'll update the PR.

mpldr: Unless you're fine with an untagged version. In that case it might be even simpler.

toml_testgen_support_test.go
```diff
@@ -1,3 +1,4 @@
+//go:generate go run github.com/toml-lang/toml-test/cmd/toml-test@master -copy ./tests
```
mpldr, line 1: This now always uses the latest available version, which might be in your interest as you can now generate tests for the new examples as soon as they come out.

cmd/tomltestgen/main.go
```diff
@@ -138,61 +139,56 @@ func main() {
...
-
 	collection := testsCollection{
 		Ref:       *ref,
 		Timestamp: time.Now().Format(time.RFC3339),
 	}
 
-	zipFilesMap := map[string]*zip.File{}
+	dirContent, _ := filepath.Glob("tests/invalid/**/*.toml")
+	for _, f := range dirContent {
+		filename := path.Base(f)
```
pelletier: I think the duplicated test issue comes from here. The previously used regular expression includes any directory after `invalid|valid` in the name, while this code only looks at the file name, discarding intermediate directories. As a result, it has the same name for `invalid/array/no-close-1` and `invalid/string/no-close-1`.

toml_testgen_support_test.go
pelletier: Yep that's cool. As long as the tests don't include examples for unreleased TOML version that's fine. But from toml-test's readme, seems like we would need to use different arguments to pull TOML 1.1 examples.

cmd/tomltestgen/main.go
mpldr: Gotcha.

mpldr:
Now all issues should be resolved.

pelletier:
Looking good! Thank you for taking care of this!

PR #928, opened 2024-01-30
Fix indentation of README examples

rdbo:
This pull request fixes the indentation in the `README.md` file. Currently it has many different non-standard indentations:
- 1 tab + 2 spaces
- 2 spaces in one line, tab in another
etc
This PR standardizes the indentation by using tabs, which is what `gofmt` uses by default.

---

Benchmarks not needed


pelletier:
Good catch, thank you!

PR #921, opened 2023-12-31
feat(unmarshaler): add support for fields to unmarshal themselves

pkarakal:
This adds the `Unmarshaler` interface to the v2 package. This way types that implement this interface can define the way they are unmarshalled. 

This is a first attempt to fix #873 

---
```
goos: linux
goarch: amd64
pkg: github.com/pelletier/go-toml/v2/benchmark
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
                                  │ /tmp/tmp.HX8I2C7g6n-v2 │       /tmp/tmp.4y6TDyGRqa-HEAD       │
                                  │         sec/op         │    sec/op     vs base                │
UnmarshalDataset/config-2                     26.76m ±  3%   29.36m ± 18%   +9.71% (p=0.000 n=10)
UnmarshalDataset/canada-2                     105.2m ±  3%   115.2m ±  2%   +9.54% (p=0.000 n=10)
UnmarshalDataset/citm_catalog-2               35.86m ± 11%   38.82m ±  8%   +8.27% (p=0.009 n=10)
UnmarshalDataset/twitter-2                    15.44m ±  4%   16.22m ±  3%   +5.01% (p=0.005 n=10)
UnmarshalDataset/code-2                       117.0m ±  1%   119.3m ±  2%   +1.98% (p=0.000 n=10)
UnmarshalDataset/example-2                    223.7µ ±  5%   236.9µ ± 10%   +5.91% (p=0.009 n=10)
Unmarshal/SimpleDocument/struct-2             776.9n ±  1%   806.3n ±  2%   +3.78% (p=0.000 n=10)
Unmarshal/SimpleDocument/map-2                1.089µ ± 13%   1.105µ ± 18%        ~ (p=0.101 n=10)
Unmarshal/ReferenceFile/struct-2              62.64µ ±  1%   66.18µ ±  1%   +5.65% (p=0.002 n=10)
Unmarshal/ReferenceFile/map-2                 97.20µ ±  5%   99.87µ ± 13%        ~ (p=0.089 n=10)
Unmarshal/HugoFrontMatter-2                   15.98µ ±  1%   16.67µ ±  1%   +4.32% (p=0.002 n=10)
Marshal/SimpleDocument/struct-2               551.2n ±  3%   548.4n ± 17%        ~ (p=0.810 n=10)
Marshal/SimpleDocument/map-2                  735.4n ±  9%   737.9n ±  1%        ~ (p=0.218 n=10)
[...]

PR #940, opened 2024-03-14
Decode: unstable/Unmarshal interface

### LINKED ISSUE #873: Bring back toml.Unmarshaler
Creating a main issue to track this work, as it has been asked multiple times, is probably hindering the adoption of go-toml v2.

The main difficulty of building this feature is modifying the unmarshaler in a way that allows it to retrieve actual bytes from the parser for the current target element without impacting performance.

The good thing is that solving this problem will also make implementing`toml.RawMessage` pretty easy (https://github.com/pelletier/go-toml/issues/796)!

In my opinion, the work should probably be carried out in this order:

1. Add the ability to retrieve the offset in the TOML document where a node's data starts. The recent merge of https://github.com/pelletier/go-toml/pull/860 introducing `Position` should help with that.
2. Make the `Decoder` detect that its current target implements `toml.Unmarshaler`, then keep calling the parser until it finishes its current structure.
3. Record the first node's start offset and the last node's end offset to retrieve the raw data from the input document.
4. Feed the raw data to the `UnmarshalTOML` method.

Part 2 is probably the most difficult to pull off; the unmarshaling and parsing code is somewhat intertwined – though thanks to the intermediate partial AST design, separating both should be doable. The hard part is keeping the same performance level when decoding non-Unmarshaler targets and trying not to fully duplicate the Decoder (otherwise, it's twice the maintenance work :)).

Attention should also be given to detecting that the type implements `toml.Unmarshaler`. There's likely going to be a performance penalty there until some parser generation is implemented (https://github.com/pelletier/go-toml/pull/669, https://github.com/pelletier/go-toml/pull/758), but we should aim to minimize the impact of that check.

rszyma:
This adds the Unmarshaler interface to the v2 package. This way types that implement this interface can define the way they are unmarshalled.

Attempts to fix #873	
Supersedes #921

### DIFF
--- a/unmarshaler.go
+++ b/unmarshaler.go
@@ -59,6 +59,14 @@ func (d *Decoder) DisallowUnknownFields() *Decoder {
 
 // EnableUnmarshalerInterface allows to enable unmarshaler interface.
 //
+// With this feature enabled, types implementing the unstable/Unmarshaler
+// interface can be decoded from any structure of the document. It allows types
+// that don't have a straightfoward TOML representation to provide their own
+// decoding logic.
+//
+// Currently, types can only decode from a single value. Tables and array tables
+// are not supported.
+//
 // *Unstable:* This method does not follow the compatibility guarantees of
 // semver. It can be changed or removed without a new major version being
 // issued.
--- a/unmarshaler.go
+++ b/unmarshaler.go
@@ -41,6 +41,9 @@ type Decoder struct {
 
 	// global settings
 	strict bool
+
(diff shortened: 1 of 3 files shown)


rszyma:
there are still arrays and tables left to be implemented properly

pelletier:
Thank you for pushing this! Sorry, it took me a while to review.

I've confirmed there is no performance regression when the feature is disabled (by default), as expected. I'll push a commit on this branch to warn people of the state of the feature and it should be good to be merged.

PR #949, opened 2024-05-03
go.mod: bump minimum language version to 1.21

mvdan:
(see commit message)


pelletier:
Thank you for submitting a PR!

My current stance on this is https://github.com/pelletier/go-toml/discussions/938. In other words, I'm not fundamentally against bumping the minimum go version, but only if there is a clear benefit to it. Since it would be preventing people stuck on older versions to use the library, the trade-off needs to be appealing.

The readme advertises that only the last two versions of Go are supported, but a best effort to keep compatibility versions seem fine to me.

Is it causing any issue or preventing you to do something on your end?

mvdan:
Interesting - I tend to take the opposite stance, I bump go.mod as soon as my minimum supported Go version goes up. Any older Go version stops being tested in CI, so for me personally, the support is already gone. Even if the code happens to still build on a slightly older Go version, I'm not testing it, so all bets are off. It's in my opinion better to signal that via a minimum version in go.mod than to leave the users in a bit of a limbo.

Nowadays there shouldn't be an excuse to stay on an older and unsupported Go version (without security updates even), particularly given https://go.dev/doc/toolchain :)

In practical terms, yes, it would be very nice to be able to use newer std additions and generics. As master is right now, we only test on Go 1.21+, but we are not able to use any standard library APIs added after 1.16, which is contradictory.

pelletier:
The performance improvement is worth the bump. Let's do it!

mimicode:
stupidity！

Issue #812, opened 2022-08-26
undefined: reflect.Pointer


cxy1620541673:
**Describe the bug**
# github.com/pelletier/go-toml/v2
D:\go\pkg\mod\github.com\pelletier\go-toml\v2@v2.0.4\unmarshaler.go:1143:37: undefined: reflect.Pointer
D:\go\pkg\mod\github.com\pelletier\go-toml\v2@v2.0.4\unmarshaler.go:1211:20: undefined: reflect.Pointer

**To Reproduce**
go get -u all

**Expected behavior**
dont report error when do "go get -u "

**Versions**
 - go-toml: v2.0.4
 - go: v1.17.11
 - operating system: Windows

**Additional context**
Add any other context about the problem here that you think may help to diagnose.

cxy1620541673:
无法更新到最新版

mohankc:
Same issue when we create a docker image.
 # github.com/pelletier/go-toml/v2
#32 287.9 ../../../../pkg/mod/github.com/pelletier/go-toml/v2@v2.0.4/unmarshaler.go:1143:37: undefined: reflect.Pointer
#32 287.9 ../../../../pkg/mod/github.com/pelletier/go-toml/v2@v2.0.4/unmarshaler.go:1211:20: undefined: reflect.Pointer

moorereason:
`reflect.Ptr` was renamed to `reflect.Pointer` in Go 1.18.  Use a newer Go release to work around this issue.

`go-toml` should use `reflect.Ptr` if we want to maintain backwards compatibility.  From the [Go 1.18 release notes](https://go.dev/doc/go1.18):

> [`reflect.Ptr`](https://go.dev/pkg/reflect#Ptr) and [`reflect.PtrTo`](https://go.dev/pkg/reflect#PtrTo) have been renamed to [`reflect.Pointer`](https://go.dev/pkg/reflect#Pointer) and [`reflect.PointerTo`](https://go.dev/pkg/reflect#PointerTo), respectively, for consistency with the rest of the reflect package.  The old names will continue to work, but will be deprecated in a future Go release.

pelletier:
I'll push a new version with `reflect.Ptr` given it's a trivial change, but keep in mind that go-toml only supports the last two versions of Go (https://github.com/pelletier/go-toml#versioning):

> The last two major versions of Go are supported (see [Go Release Policy](https://golang.org/doc/devel/release.html#policy)).

pelletier:
[...]

PR #956, opened 2024-06-25
Fix reflect.Pointer backward compatibility

xxxVitoxxx:

```
/go/pkg/mod/github.com/pelletier/go-toml/v2@v2.2.2/marshaler.go:735:40: undefined: reflect.Pointer
```

If we want to reflect.Pointer backward compatibility. here should use reflect.Ptr.

related issue: https://github.com/pelletier/go-toml/issues/812

---

Benchmarks not needed


pelletier:
Thank you!

PR #958, opened 2024-07-16
Allow `int`, `uint`, and `float` as map keys

daniel-weisse:
According to the [toml specification](https://toml.io/en/v1.0.0#keys), integers (or even floats) may be used as keys.
This PR adds functionality to `*Encoder.keyToString` and `*decoder.keyFromData` to convert `int`s,  `uint`s, and `float`s to and from string.

-----------------------

[benchstat results](https://github.com/user-attachments/files/16249393/benchstat.txt)


pelletier:
Feels like a good ergonomics improvement. Thank you for the patch!

PR #959, opened 2024-08-07
Update goreleaser action to `v6` and set goreleaser binary to `v2`

daniel-weisse:
`goreleaser` recently released version 2 of the tool.
This release deprecates the `--rm-dist` flag.
Since the `version` argument of the `goreleaser/goreleaser-action` of the `release.yml` workflow was set to `latest`, the action downloaded the newer version, breaking the workflow.

This PR updates the goreleaser action to `v6` and pins the goreleaser binary version to any minor version of `v2`.
This should save the workflow from running into similar issues if gorelaser ever releases a `v3` with other breaking changes.

b2e0fda3392425cc434eb42551509347e064322e updates the `workflow.yml` workflow to use the `release.yml` from the same commit the workflow was triggered on.
This is required for the checks to pass on this PR, as it updates that very same workflow.
Otherwise it takes the workflow from the last tag on `v2`.
If this is desired behavior, I can just drop the commit before merging.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,10 +30,10 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v3
+        uses: goreleaser/goreleaser-action@v6
         with:
           distribution: goreleaser
-          version: latest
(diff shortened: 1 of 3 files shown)


pelletier:
Thank you for taking care of this! The change looks good to me, I will try to release a new version of go-toml sometime this coming week and we'll see how it goes!

PR #967, opened 2024-09-19
support time.Duration

finchcn:

support parse string as time.Duration 

---


finchcn:
Hi  @pelletier , could you please check these changes and give some feedback?

pelletier:
Hi! Sorry for the delay.

Thank you for the patch! However I don't think duration parsing has it's place in a toml library (unless the TOML spec defines durations as a type). Instead a higher "config" level would be better suited for that. You should be able to achieve the same effect by wrapping time.Duration in a type that implement TextUnmarshaler.

PR #969, opened 2024-10-11
Allow tranforming LocalTime to time.Duration

Benbebop:
Adds a function to transform LocalTime to time.Duration and does it automatically during unmarshalling.

[benchstat.txt](https://github.com/user-attachments/files/17337319/benchstat.txt)

### DIFF
--- a/localtime.go
+++ b/localtime.go
@@ -50,6 +50,11 @@ type LocalTime struct {
 	Precision  int // Number of digits to display for Nanosecond.
 }
 
+// AsDuration converts d into a duration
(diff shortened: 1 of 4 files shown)


pelletier:
Hi! Sorry for the delay.

Thank you for the PR! Similarly to https://github.com/pelletier/go-toml/pull/967#issuecomment-2424136823, I don't think duration support should be added to the library itself, but rather on a higher-level library. Also here it changes the meaning of local time to duration. Since `LocalTime.AsDuration` only uses public fields, so it can be implemented outside of the library.

PR #989, opened 2025-05-30
Use reflect.Value.IsZero for omitempty checks

NathanBaulch:
Struct fields of type `time.Time` and `netip.Addr` are always omitted when using `omitempty` because the current `isEmptyValue` function can only check structs with exported fields. See #955 and #988.

Replacing the current `isEmptyValue` function with `reflect.Value.IsZero()` is a very simple and low risk fix that should make `omitempty` useful for many more types.

---

```text
goos: linux
goarch: amd64
pkg: github.com/pelletier/go-toml/v2/benchmark
cpu: Intel(R) Core(TM) Ultra 9 275HX
                                   │ old_bench_10.txt │          new_bench_10.txt          │
                                   │      sec/op      │   sec/op     vs base               │
UnmarshalDataset/config-24                11.95m ± 3%   12.28m ± 3%       ~ (p=0.105 n=10)
UnmarshalDataset/canada-24                50.90m ± 6%   46.93m ± 3%  -7.80% (p=0.002 n=10)
UnmarshalDataset/citm_catalog-24          21.77m ± 4%   21.27m ± 1%  -2.31% (p=0.035 n=10)
UnmarshalDataset/twitter-24               7.624m ± 6%   8.092m ± 4%  +6.15% (p=0.029 n=10)
UnmarshalDataset/code-24                  51.21m ± 6%   50.42m ± 4%       ~ (p=0.165 n=10)
UnmarshalDataset/example-24               111.4µ ± 3%   115.9µ ± 5%       ~ (p=0.089 n=10)
Unmarshal/SimpleDocument/struct-24        354.5n ± 3%   380.2n ± 3%  +7.25% (p=0.000 n=10)
Unmarshal/SimpleDocument/map-24           494.8n ± 4%   517.0n ± 4%  +4.50% (p=0.029 n=10)
Unmarshal/ReferenceFile/struct-24         27.74µ ± 5%   29.04µ ± 5%  +4.68% (p=0.035 n=10)
Unmarshal/ReferenceFile/map-24            42.84µ ± 5%   42.85µ ± 4%       ~ (p=0.853 n=10)
Unmarshal/HugoFrontMatter-24              7.200µ ± 3%   7.374µ ± 3%       ~ (p=0.143 n=10)
Marshal/SimpleDocument/struct-24          258.1n ± 6%   263.9n ± 3%       ~ (p=0.271 n=10)
[...]

Issue #988, opened 2025-05-17
CreatedAt, UpdatedAt are skipped, when toml.Marshal


snowdream:
**Describe the bug**

```
type User struct {
	// Model embeds common database fields (ID, CreatedAt, UpdatedAt, DeletedAt)
	// that are shared across most entity types in the system
	// These fields automatically track creation, update, and deletion timestamps
	Model // Embedded base model with ID and timestamp fields (created_at, updated_at, deleted_at)

	// UserName is the unique identifier for the user in the system.
	// It must not be empty and is indexed in the database.
	// This field is used for login purposes and must be unique across all users.
	// Validation ensures this field is never empty before persistence.
	// Users will use this value to identify themselves during authentication.
	// The uniqueness constraint prevents duplicate registrations and ensures system integrity.
	UserName string `json:"username" xml:"username" yaml:"username" toml:"username" schema:"username"  gorm:"column:username;not null;uniqueIndex:user_ids;default:'';"`
...
```

```
// Model represents the base structure that other domain models can embed.
// It provides standard fields for identification, creation time, update time,
// and soft deletion support through GORM's DeletedAt field.
type Model struct {
	// ID is the primary key for the model record in the database.
	ID id.ID `json:"id" xml:"id" yaml:"id" toml:"id" schema:"id" gorm:"primaryKey;autoIncrement;column:id"`
	// CreatedAt stores the timestamp when the record was initially created.
	CreatedAt time.Time `json:"created_at,omitempty" xml:"created_at,omitempty"  yaml:"created_at,omitempty" toml:"created_at,omitempty" schema:"created_at,omitempty" gorm:"column:created_at"`
	// UpdatedAt stores the timestamp when the record was last updated.
	UpdatedAt time.Time `json:"updated_at,omitempty" xml:"updated_at,omitempty" yaml:"updated_at,omitempty"  toml:"updated_at,omitempty" schema:"updated_at,omitempty" gorm:"column:updated_at"`
	// DeletedAt supports soft deletion (when set, the record is considered deleted).
	DeletedAt gorm.DeletedAt `json:"deleted_at,omitempty" xml:"deleted_at,omitempty"  yaml:"deleted_at,omitempty" toml:"deleted_at,omitempty" schema:"deleted_at,omitempty" gorm:"index;column:deleted_at"`
}
```

**To Reproduce**
when i try to Marshal with toml, CreatedAt, UpdatedAt are skipped.
CreatedAt, UpdatedAt should be output after id.
```
[...]

PR #998, opened 2025-08-22
Add `omitzero` tag support

NathanBaulch:
As requested in #989 and mentioned in #988. This is a simple non-breaking additive marshaler change that introduces `omitzero` alongside `omitempty`. Fields decorated with this tag will be omitted if `reflect.Value.IsZero()` returns true.


pelletier:
Thanks! It needs a couple more tests to validate it works correctly with various tables (likely why the coverage report is failing).

NathanBaulch:
Done!

PR #1012, opened 2025-12-19
Unwrap strict errors

bersace:
`Unwrap() error` and `Unwrap() []error` are common API implemented by standard `errors` package.

I suggest to simply implement `Unwrap() []error` on `StrictMissingError()` to ease handling of wrapped decode errors.

What do you think of this ?


pelletier:
Seems like a good idea! Never realized that’s something you could provide.

PR #1013, opened 2025-12-21
Use constant format strings with Printf-like functions

flyn-org:
Recent versions of Go object to the use of non-constant variables a format strings. This commit fixes errors like this:

cli.go:26:47: non-constant format string in call to fmt.Fprintf

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -22,7 +22,7 @@ type Program struct {
 }
 
 func (p *Program) Execute() {
-	flag.Usage = func() { fmt.Fprintf(os.Stderr, p.Usage) }
+	flag.Usage = func() { fmt.Fprint(os.Stderr, p.Usage) }
 	flag.Parse()
 	os.Exit(p.main(flag.Args(), os.Stdin, os.Stdout, os.Stderr))
 }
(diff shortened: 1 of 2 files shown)


internal/cli/cli.go
```diff
@@ -22,7 +22,7 @@ type Program struct {
 }
 
 func (p *Program) Execute() {
-	flag.Usage = func() { fmt.Fprintf(os.Stderr, p.Usage) }
+	flag.Usage = func() { fmt.Fprintf(os.Stderr, "%s", p.Usage) }
```
ccoVeille: This is OK, but you could have [fmt.Fprint](https://pkg.go.dev/fmt#Fprint) (sane one without -f)  as you need no formatting


```suggestion
	flag.Usage = func() { fmt.Fprint(os.Stderr, p.Usage) }
```

ccoVeille (approved):

internal/cli/cli.go
flyn-org: Good point. I just modified the patch to use Fprint.

ccoVeille (approved):

pelletier (approved):

PR #1016, opened 2025-12-31
Improve Unmarshaling README

heckelson:

- I clarified that go struct vars should be capitalized and toml variables don't necessarily have to be. 
- I added a section describing how to unmarshal a (nested) toml table.

More detailed description of the decisions being made and the reasons why (if
the patch is non-trivial).

- For the first point, it took me a while to understand that my capitalization was the issue.
- To accomplish the second part, I had to look into the Gitlab discussions to find an example on how to parse nested documents. It would be better if we had an example in the readme IMO.

---

N/A, since this is just a documentation change

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,11 @@ type MyConfig struct {
 ### Unmarshaling
 
 [`Unmarshal`][unmarshal] reads a TOML document and fills a Go structure with its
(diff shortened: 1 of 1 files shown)


pelletier:
Thanks for adding the clarification!

PR #1020, opened 2026-01-04
Support custom IsZero() methods with omitzero tag

### LINKED ISSUE #1003: Custom IsZero() method on type is ignored when using omitzero tag
**Describe the bug**
When using the `omitzero` struct tag, the current implementation calls `reflect.Value.IsZero`
 to determine whether a value should be omitted. However, this ignores user-defined `IsZero()` method that some types implement (similar to how encoding/json checks for them).

https://github.com/pelletier/go-toml/blob/a0e846496753c39d856894fe5ee82fd9df0f14f1/marshaler.go#L390-L392

**To Reproduce**
```go
package main

import (
	"fmt"
	"log"

)

type Custom struct {
	Value int
}

func (c Custom) IsZero() bool {
	return c.Value < 10
}

type Example struct {
	Field Custom `toml:",omitzero"`
}

func main() {
	data := Example{Field: Custom{Value: 5}}

	d, err := toml.Marshal(data)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%s\n", string(d))
}

```

**Expected behavior**
Field should be omitted because `Custom.IsZero()` returns true.

**Versions**
 - go-toml: a0e846496753c39d856894fe5ee82fd9df0f14f1
 - go: version 1.24.6
 - operating system: Windows, Linux

pelletier:
The omitzero tag now respects custom IsZero() methods on types,
similar to how encoding/json handles this. Previously, only
reflect.Value.IsZero() was used, which ignores user-defined
implementations.

Fixes #1003

### DIFF
--- a/marshaler.go
+++ b/marshaler.go
@@ -390,7 +390,28 @@ func shouldOmitEmpty(options valueOptions, v reflect.Value) bool {
 }
 
 func shouldOmitZero(options valueOptions, v reflect.Value) bool {
-	return options.omitzero && v.IsZero()
+	if !options.omitzero {
+		return false
+	}
+
+	// Check if the type implements isZeroer interface (has a custom IsZero method).
+	if v.Type().Implements(isZeroerType) {
+		return v.Interface().(isZeroer).IsZero()
+	}
+
+	// Check if pointer type implements isZeroer.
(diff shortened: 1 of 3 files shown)


pelletier:
@claude resolve conflicts.

PR #1048, opened 2026-04-08
README.md: remove reference to old go versions and modules

maruel:
go.mod requires 1.21. We're long past 1.13, so there's no point in having reference to go modules in the README.md, this is just confusing for the readers.

No code change.


pelletier:
I was initially going to say that i added this not too long ago. It's been 5 years.

PR #1050, opened 2026-04-10
Change "missing field" to "unknown field" in strict mode errors

phinze:
When `DisallowUnknownFields` is enabled and the TOML document contains a key
that doesn't match a struct field, the diagnostic annotation says "missing
field". This reads as though the TOML file is lacking something, when really it
contains something the decoder doesn't recognize.

This changes the user-facing message from "missing field" to "unknown field",
which is consistent with the `DisallowUnknownFields` API name and describes the
problem from the user's point of view.

The `StrictMissingError` type name is left unchanged to avoid a breaking API
change.

---

No performance-sensitive code paths are changed (string literals only), so
benchstat is not applicable.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -24,9 +24,9 @@ type DecodeError struct {
 	human string
 }
 
-// StrictMissingError occurs in a TOML document that does not have a
-// corresponding field in the target value. It contains all the missing fields
-// in Errors.
+// StrictMissingError occurs when the TOML document contains fields that do not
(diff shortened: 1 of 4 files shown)


phinze:
Hi @pelletier and @Yanhu007 not sure why you re-implemented this in #1054 a few days later when I linked it from the original issue. Maybe an oversight? It would have been nice to contribute, but oh well, glad it's fixed!

PR #1051, opened 2026-04-12
Fix parser error position after comment lines

### LINKED ISSUE #1047: v2.3.0 regression with "invalid character at start of key" error positions
```
package main

import (
	"testing"

)

	p := toml.Parser{}
	p.Reset(input)
	for p.NextExpression() {
	}
	err := p.Error()
	if err == nil {
		t.Fatal("expected an error")
	}
	t.Logf("Error: %v", err)
	r := p.Range(perr.Highlight)
	shape := p.Shape(r)

	if r.Offset != 10 {
	}
	}
}
```

lawrence3699:
## Summary

Fix `unstable.Parser` reporting wrong error positions when the error occurs after a comment line.

`subsliceOffset` assumed highlight slices were always suffixes of the input, using `len(data) - len(b)` to compute the offset. After comment lines are skipped, error highlights are interior subslices, not suffixes, causing incorrect positions.

Fixes #1047

## Error position example

Given input `# comment\n= "value"`:

**Before** — `=` is at byte 10 (line 2, col 1), but reported at byte 18 (line 2, col 9)

**After** — correctly reported at byte 10 (line 2, col 1)

## Fix

Use pointer comparison via `reflect.ValueOf().Pointer()`, matching the approach already used in `errors.go`'s `subsliceOffset`.

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] New regression test `TestParser_ErrorPositionAfterComment` verifies correct offset and position

### DIFF
--- a/unstable/parser.go
+++ b/unstable/parser.go
@@ -3,6 +3,7 @@ package unstable
 import (
 	"bytes"
 	"fmt"
+	"reflect"
 	"unicode"
 
 	"github.com/pelletier/go-toml/v2/internal/characters"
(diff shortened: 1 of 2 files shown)


pelletier:
Should be fixed by https://github.com/pelletier/go-toml/pull/1056

PR #1052, opened 2026-04-13
fix: omitempty should not treat TextMarshaler types as empty

### LINKED ISSUE #955: omitempty ignores types implementing TextMarshaler
**Describe the bug**

**To Reproduce**
Minimal example:
```go
	tmp := struct {
	}{
	}
	if err != nil {
		t.Fatal(err)
	}
	}
}
```

**Expected behavior**

**Versions**
 - go-toml: 2.2.2
 - go: 1.22.4
 - operating system: Linux

Yanhu007:
Fixes #955

## Problem

Types like `netip.Addr` implement `encoding.TextMarshaler` but have no exported fields. `isEmptyValue` checks struct emptiness by iterating exported fields, so these types are always treated as empty:

```go
type Config struct {
}

// netip.MustParseAddr("192.168.1.1") → omitted! (no exported fields = "empty")
```

## Fix

1. Check `isZeroer` interface first (types with custom `IsZero()` like `time.Time`)
2. For `TextMarshaler` types, use `reflect.IsZero` instead of field-based check
3. Fall back to existing field-based check for regular structs

This ensures that `netip.Addr`, `uuid.UUID`, and other TextMarshaler types are only omitted when they are truly zero-valued.

All existing tests pass.

### DIFF
--- a/marshaler.go
+++ b/marshaler.go
@@ -449,8 +449,21 @@ func (enc *Encoder) commented(commented bool, b []byte) []byte {
 }
 
 func isEmptyValue(v reflect.Value) bool {
+	// For types that implement isZeroer (custom IsZero), defer to that.
+	if v.Type().Implements(isZeroerType) {
+		return v.Interface().(isZeroer).IsZero()
+	}
+	if v.CanAddr() && reflect.PointerTo(v.Type()).Implements(isZeroerType) {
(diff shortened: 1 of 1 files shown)


pelletier:
Was fixed recently, sorry.

PR #1053, opened 2026-04-13
marshaler: respect TextMarshaler and IsZero in shouldOmitEmpty

### LINKED ISSUE #955: omitempty ignores types implementing TextMarshaler

alliasgher:
## Summary

Fixes #955.

`isEmptyStruct` scans exported struct fields; a struct with no exported fields (e.g. `netip.Addr`) was always considered empty, so any such struct field tagged `omitempty` was always omitted even when it held a meaningful value.

### Root cause

`isEmptyStruct` iterates over exported fields. `netip.Addr` has none, so the loop exits immediately and returns `true` (empty) regardless of the addr's value.

### Fix

`shouldOmitEmpty` is updated to mirror the logic already in `shouldOmitZero`:

1. **`IsZero()` wins** — if the type (or its pointer) implements the `isZeroer` interface, use that, exactly as `shouldOmitZero` does.
2. **TextMarshaler fallback** — if the type is a struct that implements `encoding.TextMarshaler` but not `isZeroer`, fall back to `reflect.Value.IsZero()` rather than the exported-field scan.

### Before

```go
// netip.Addr{"192.168.178.35"} is silently dropped — ip key absent
```

### After

```go
// ip = '192.168.178.35'  — correctly emitted
```

## Tests

- `TestEncoderOmitemptyTextMarshalerNonZero` — non-zero `netip.Addr` with `omitempty` is emitted
- `TestEncoderOmitemptyTextMarshalerZero` — zero `netip.Addr` is still omitted
- Full `go test ./...` passes

### DIFF
--- a/marshaler.go
+++ b/marshaler.go
@@ -390,32 +390,27 @@ func shouldOmitEmpty(options valueOptions, v reflect.Value) bool {
 		return false
 	}
 
-	// If the value implements IsZero() use it, like shouldOmitZero does.
-	// This is important for types such as netip.Addr that have no exported
-	// fields (making isEmptyStruct incorrectly report them as empty) but do
-	// implement a meaningful zero check.
-	if v.Type().Implements(isZeroerType) {
-		return v.Interface().(isZeroer).IsZero()
-	}
-	if reflect.PointerTo(v.Type()).Implements(isZeroerType) {
-		if v.CanAddr() {
-			return v.Addr().Interface().(isZeroer).IsZero()
-		}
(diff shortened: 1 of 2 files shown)


pelletier:
Was fixed recently, sorry.

PR #1055, opened 2026-04-14
unstable: fix subsliceOffset regression for non-suffix subslices

### LINKED ISSUE #1047: v2.3.0 regression with "invalid character at start of key" error positions

alliasgher:
Fixes #1047.

`subsliceOffset` computed byte offsets by subtracting the slice length from `len(p.data)`. This only works for **tail (suffix)** slices. Sub-slices taken from the middle — e.g. `b[0:1]` in `parseSimpleKey`'s error path — return a wrong offset.

This was exposed by #1041 (which correctly fixed a different bug): after that change the `parseSimpleKey` error highlight is still `b[0:1]`, a non-suffix slice, so the offset computation gives `len(p.data) - 1` instead of the actual character position.

**Fix:** use pointer arithmetic (`unsafe.Pointer`) to compute the actual offset of any subslice relative to `p.data`. An empty-slice fallback preserves the existing behaviour for nil inputs.

Includes `TestRangeOffsetAfterComment` from #1047.

### DIFF
--- a/unstable/parser.go
+++ b/unstable/parser.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"unicode"
+	"unsafe"
 
 	"github.com/pelletier/go-toml/v2/internal/characters"
 )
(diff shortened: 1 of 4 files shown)


unstable/parser.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"unicode"
+	"unsafe"
```
pelletier: Sorry I don't want to re introduce unsafe.

pelletier (changes requested):

unstable/parser.go
alliasgher: Replaced with `cap(p.data) - cap(b)` — no unsafe import needed.

pelletier:
Merged https://github.com/pelletier/go-toml/commit/c1712164d472b41628043ceb48d5f5da34c3922b to solve this with more testing. I think there is a deeper fix to be made but good enough to fix the symptoms for users.

PR #1056, opened 2026-04-16
Fix incorrect error positions in unstable parser Range() (#1047)

### LINKED ISSUE #1047: v2.3.0 regression with "invalid character at start of key" error positions

pelletier:
Parser.subsliceOffset() assumed its input was always a suffix of the
document, computing offset as len(data)-len(b). Error highlights from
NewParserError are arbitrary subslices (e.g. b[0:1]), not suffixes,
causing wrong offsets after comments and other consumed content.

Replace the suffix-length calculation with reflect-based pointer
arithmetic, matching the safe approach already used in errors.go.

Fixes #1047

### DIFF
--- a/errors_test.go
+++ b/errors_test.go
@@ -301,6 +301,123 @@ OtherMissing = 1
 	assert.Equal(t, 2, len(strictErr.Unwrap()))
 }
 
+func TestDecodeError_PositionAfterComments(t *testing.T) {
+	examples := []struct {
+		name        string
+		doc         string
+		expectedRow int
+		expectedCol int
+	}{
+		{
+			name:        "comment then invalid key start",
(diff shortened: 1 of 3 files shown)


pelletier:
I don't love that change. I think there is probably a deeper change to be made to explicitly track locations while parsing. But good enough to fix the current bug.

PR #1059, opened 2026-05-09
marshaler: respect TextMarshaler and IsZero() in omitempty struct check

### LINKED ISSUE #955: omitempty ignores types implementing TextMarshaler

nghiack7:
## Problem

Fixes #955

\`omitempty\` silently drops struct values that implement \`encoding.TextMarshaler\` but have no exported fields (e.g. \`netip.Addr\`, \`netip.Prefix\`, custom wrapper types).

Root cause: \`isEmptyStruct\` only walks **exported fields** to decide emptiness. A type like \`netip.Addr\` has no exported fields, so it always returns \`true\` → the field is skipped regardless of the actual value.

## Fix

Before the exported-field walk, \`isEmptyStruct\` now checks:

1. **\`IsZero() bool\` method** — if the type implements the \`isZeroer\` interface, delegate directly to it
2. **\`encoding.TextMarshaler\`** — if the type controls its own serialization, treat the value as non-empty

This matches the approach already used by \`shouldOmitZero\`.

## Tests

Added \`TestEncoderOmitemptyTextMarshaler\` covering \`netip.Addr\` with \`omitempty\`.

### DIFF
--- a/marshaler.go
+++ b/marshaler.go
@@ -470,6 +470,29 @@ func isEmptyValue(v reflect.Value) bool {
 }
 
 func isEmptyStruct(v reflect.Value) bool {
+	// If the type has an IsZero() bool method, delegate to it.
+	// This covers types like netip.Addr that have no exported fields but carry
+	// meaningful state (e.g. a non-zero address should not be omitted).
+	if v.Type().Implements(isZeroerType) {
+		return v.MethodByName("IsZero").Call(nil)[0].Bool()
+	}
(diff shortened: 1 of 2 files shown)


pelletier:
Was fixed recently, sorry.

PR #1060, opened 2026-05-16
marshaler: respect TextMarshaler when checking omitempty on structs

### LINKED ISSUE #955: omitempty ignores types implementing TextMarshaler

c-tonneslan:
Fixes #955.

`netip.Addr` (and other TextMarshaler-only types like `time.Time`, `big.Int`) has no exported fields, so `isEmptyStruct` walks an empty list and always reports the value as empty. With `omitempty` set on a field of that type, the value silently gets dropped from the output:

```go
type doc struct {
    IP netip.Addr `toml:\"ip,omitempty\"`
}
toml.Marshal(doc{IP: netip.MustParseAddr(\"192.168.178.35\")})
// got:  \"\"
// want: \"ip = '192.168.178.35'\\n\"
```

Check for `encoding.TextMarshaler` (on the value or its pointer) before falling back to the field walk. If MarshalText returns nothing, the value is considered empty; otherwise it's kept. The existing field walk handles the rest unchanged.

Targeting the `v2` branch.

Added `TestEncoderOmitempty_TextMarshaler` with three cases: a non-zero netip.Addr is kept, a zero netip.Addr is omitted, and a non-zero time.Time is kept.

### DIFF
--- a/marshaler.go
+++ b/marshaler.go
@@ -470,6 +470,19 @@ func isEmptyValue(v reflect.Value) bool {
 }
 
 func isEmptyStruct(v reflect.Value) bool {
+	// A struct that marshals via TextMarshaler (netip.Addr, time.Time, ...)
+	// has no exported fields the field-walk below can see, so it would
+	// always look empty. Defer to MarshalText and treat a zero-length
+	// result as empty.
+	if v.Type().Implements(textMarshalerType) {
+		text, err := v.Interface().(encoding.TextMarshaler).MarshalText()
+		return err == nil && len(text) == 0
+	}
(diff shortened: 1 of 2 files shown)


pelletier:
Was fixed recently, sorry.

PR #1063, opened 2026-05-25
Report quoted string requirement for unquoted values

### LINKED ISSUE #413: missing quotes gives misleading error message
**Describe the bug**
The error message for missing quotes is misleading.

**To Reproduce**
1. put something like `timeout_broadcast_tx_commit = 20s`
2. start some utility that uses this library (such as cosmos-sdk)
3. `ERROR: While parsing config: (136, 33): parsing error: keys cannot contain new lines`

**Expected behavior**
`ERROR: While parsing config: (136, 33): parsing error: value requires quotes` or some such

**Versions**
 - go-toml: current, I think: 16c9a8b
 - go: go1.13.4.linux-amd64.tar.gz
 - operating system: Linux

wyf027:
## Summary

Fixes #413

Unquoted values with alphabetic suffixes (e.g. `timeout = 20s`) were partially parsed as numbers and then failed with a misleading `expected newline` error. The parser now rejects alphabetic characters while scanning numeric values and returns `strings must be quoted`.

## Test plan

- [x] `go test ./...`
- [x] Added `TestUnquotedStringValueRequiresQuotes` for `timeout = 20s`

### DIFF
--- a/unstable/parser.go
+++ b/unstable/parser.go
@@ -1221,6 +1221,10 @@ func (p *Parser) scanIntOrFloat(b []byte) (reference, []byte, error) {
 			return invalidReference, nil, NewParserError(b[i:i+1], "unexpected character 'n' while scanning for a number")
 		}
 
+		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') {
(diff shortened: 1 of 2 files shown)


pelletier:
Thank you!

PR #1065, opened 2026-05-25
fix: wrap redefinition errors as DecodeError with key context

### LINKED ISSUE #668: Contextualize redefinition errors
Follow up to #667. When SeenTracker finds a duplicate in the document structure it returns a simple `error` with no context. For example:

```
toml: key b is already defined
```

We should also provide more context about where the error occurred in the document, maybe the current path to the key at fault, possibly the source of the error (like where was the original key that just got redefined), similar to `DecodeError` (maybe should just reuse it?).

wyf027:
## Summary
- Wrap `SeenTracker` errors from `CheckExpression` as `DecodeError` via `decoder.wrapSeenError`
- Attach the full TOML key path and source location to duplicate key / table redefinition errors

Fixes #668

## Test plan
- [x] `go test ./...`
- [x] Added `TestDecodeErrorRedefinition` for duplicate top-level and nested keys
- [x] `b = 2` followed by `b = 3` now returns `DecodeError` with `Key() == ["b"]` and a highlighted line

### DIFF
--- a/errors_test.go
+++ b/errors_test.go
@@ -418,6 +418,31 @@ func TestErrorPositionConsistency(t *testing.T) {
 	}
 }
 
+func TestDecodeErrorRedefinition(t *testing.T) {
+	t.Run("duplicate top-level key", func(t *testing.T) {
+		m := map[string]interface{}{}
+		err := Unmarshal([]byte("a = 1\nb = 2\nb = 3\n"), &m)
+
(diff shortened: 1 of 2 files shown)


pelletier:
Pushed this over the finish line with https://github.com/pelletier/go-toml/pull/1069; thank you for your help!

PR #1066, opened 2026-05-26
fix: support encoding.BinaryUnmarshaler during decode

### LINKED ISSUE #1058: Request: Unmarshalling of *url.URL and other UnmarshalBinary types?
Some stdlib types like `*net/url.URL` implement `UnmarshalBinary` even though they operate on strings. This was necessary because adding new `UnmarshalText` representations had the risk of breaking existing programs (see golang/go#25705 and golang/go#10275).

Given that there are quite a few such types, should the TOML parser accept `UnmarshalBinary` as well as `UnmarshalText`? Or should more types be special-cased like `time` types are today?

At the moment you need to write wrapping types, which are just generally more annoying to work with (it either leads to endless casts or struct embedding that is similarly verbose).

wyf027:
## Summary

Fixes #1058

Some stdlib types such as `*url.URL` implement `encoding.BinaryUnmarshaler` but not `encoding.TextUnmarshaler`. Decoding into those types previously failed with a string type mismatch.

This adds `BinaryUnmarshaler` support for scalar values (after the existing `TextUnmarshaler` path) and for map keys.

## Test plan

- [x] Added `TestIssue1058_UnmarshalURL` and `TestIssue1058_UnmarshalURLValue`
- [x] `go test ./...`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/issue1058_test.go
+++ b/issue1058_test.go
@@ -1,0 +1,43 @@
+package toml_test
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/pelletier/go-toml/v2"
+)
+
+func TestIssue1058_UnmarshalURL(t *testing.T) {
+	type config struct {
+		Endpoint *url.URL `toml:"endpoint"`
+	}
+
+	var cfg config
+	err := toml.Unmarshal([]byte(`endpoint = "https://example.com/path"`), &cfg)
+	if err != nil {
+		t.Fatalf("Unmarshal() error = %v", err)
+	}
+
+	if cfg.Endpoint == nil {
+		t.Fatal("expected endpoint URL to be set")
+	}
+	if got, want := cfg.Endpoint.String(), "https://example.com/path"; got != want {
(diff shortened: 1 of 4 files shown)


wyf027:
Addressed the BinaryUnmarshaler map-key panic review. `keyFromData` now avoids calling `Elem()` on non-pointer key types, and there is a regression test for value-type `encoding.BinaryUnmarshaler` map keys returning an error instead of panicking. Current checks are clean.

pelletier:
Thank you for submitting this! Holding until I get a good understanding of the need at https://github.com/pelletier/go-toml/issues/1058, since it involves increasing the public API.

wyf027:
Updated this PR on top of the latest `v2` branch and covered the new BinaryUnmarshaler decode/map-key paths, including error cases.

Local validation:
- `go test ./...`
- `./ci.sh coverage -d upstream/v2` -> `Results: upstream/v2 97.6% HEAD 97.6%`, `Delta: 0`

GitHub checks have been retriggered on the updated branch.

PR #1085, opened 2026-06-28
perf: avoid interface boxing when decoding date/time values

nishantmehta:
### What

Unmarshaling a TOML date/time into a `time.Time`, `LocalDate`, `LocalTime` or `LocalDateTime` field used `v.Set(reflect.ValueOf(x))`, which boxes the value into an interface and forces a heap allocation (24 B for `time.Time`). Every other scalar assignment path uses a typed setter (`SetString`/`SetInt`/`SetFloat`) and does not allocate; date/time was the exception because `reflect` has no typed setter for struct values.

This adds a small generic helper that writes directly through the field's address when it is addressable (struct fields, slice elements), falling back to the reflect path for non-addressable targets (map elements). It removes one allocation per date/time value decoded into a struct/slice and is ~3x faster at the assignment step.

### Benchmark

```
benchmark/BenchmarkUnmarshal/ReferenceFile/struct:  88 -> 85 allocs/op,  3.424 -> 3.353 KiB/op
```

Microbenchmark of the assignment alone: `24 B/op, 1 alloc -> 0 B/op, 0 allocs` (~3x faster).

### Testing

Full test suite passes. No behavioral change — the `time.Time`/local-date values produced are identical; only the assignment mechanism changes.

### DIFF
--- a/unmarshaler.go
+++ b/unmarshaler.go
@@ -1835,17 +1835,14 @@ func (d *decoder) assignDateTime(v reflect.Value, value *unstable.Node) (reflect
 	return reflect.Value{}, d.typeMismatchError("datetime", v.Type(), d.p.Raw(value.Raw))
 }
 
-// setConcrete assigns x to v, avoiding the heap allocation that
-// v.Set(reflect.ValueOf(x)) incurs from boxing x into an interface. The fast
-// path requires v to be addressable and of x's concrete type, which holds for
-// struct fields and slice elements; map elements are not addressable, so it
-// falls back to the reflect path for them.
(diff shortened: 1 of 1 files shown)


pelletier:
Thanks, this is a nice win! The coverage gate was failing because the `!CanAddr()` fallback in `setConcrete` is unreachable: everything the decoder assigns into is settable, and in `reflect` settable implies addressable (map elements are decoded into `reflect.New`-allocated temporaries, see `descend`). A non-addressable value would have panicked in `v.Set` just the same, so the fallback added no safety — I pushed a commit removing it and updating the comment. Merging once CI is green.

PR #1084, opened 2026-06-24
Add RawMessage marshal support to unstable

pelletier:
## What

Makes `unstable.RawMessage` round-trip like `json.RawMessage` by adding the **encode** half of the feature.

The decode side already existed: `unstable.RawMessage` captures raw TOML when `Decoder.EnableUnmarshalerInterface()` is set (the `unstable.Unmarshaler` work from #873 → #940 → #1027, plus root documents in #1083). But marshaling a `RawMessage` emitted the underlying `[]byte` as an array of integers (`plugin = [110, 97, ...]`), so it could not round-trip. This PR adds the missing encoder support.

Fixes #796.

## API (all new public API stays in `unstable`)

- **`unstable.Marshaler`** — `MarshalTOML() ([]byte, error)`, the encode counterpart of `unstable.Unmarshaler`, implemented by `RawMessage` (value receiver, mirroring `json.RawMessage`). It can be called more than once for the same value during a single encode, so it must be deterministic.
- **`Encoder.EnableMarshalerInterface()`** — opt-in toggle in the `toml` package, exactly mirroring the existing `Decoder.EnableUnmarshalerInterface()` (both carry the `*Unstable:*` disclaimer). It is the only addition to the stable package, is backward compatible, and the default `toml.Marshal`/`Encode` behavior is unchanged.

## Behavior

With the interface enabled, the encoder splices a `Marshaler`'s bytes verbatim, parsing them to pick the valid position:

| RawMessage content | emitted as |
|---|---|
| a single value (`42`, `"x"`, `[1,2]`, `{a=1}`) | `key = <raw>` |
| key-value lines (`name = "x"\n...`) | `[key]` table body |
| `[]RawMessage` of table bodies | `[[key]]` array of tables |
| at the document root | the whole document (counterpart of #1083's root decode) |
| empty / whitespace | omitted |
| not valid TOML for its position | error (like `json.RawMessage` with invalid JSON) |
| table content forced inline (e.g. `SetTablesInline`) | error |

Spliced bytes are validated by the same parser the decoder uses, so a `Marshaler` cannot silently corrupt the document; validation runs against the bytes actually emitted, so even a non-deterministic implementation cannot splice table content into a value position. A table-shaped body under a `commented` field is comment-prefixed line by line, like structurally encoded tables.

```go
type Config struct {
    Plugin unstable.RawMessage `toml:"plugin"`
}

var cfg Config
toml.NewDecoder(r).EnableUnmarshalerInterface().Decode(&cfg)
[...]

PR #1088, opened 2026-07-03
Performance: tracker indexing, exact wide floats, fused decoding, encoder fast paths

pelletier:
# Performance: tracker indexing, exact wide floats, fused decoding, encoder fast paths

A systematic performance pass over the decoder and encoder — **pure safe Go
throughout** (an earlier iteration used `unsafe` slab allocation; it was
removed per the project rule, along with a chunk allocator whose shared
backing arrays could retain earlier documents' values). Measured with
interleaved A/B benchstat runs against `v2` @ 21f8286 on linux/amd64
(Threadripper 3975WX, pinned cores, n=12) and darwin/arm64 (M1 Pro, n=16),
go1.26.4.

**Results (benchstat medians, candidate @ daad458, baseline v2 @ 57fec25 —
the current v2 HEAD, including its own date/time perf fix):** geomean
**1.35× on linux, 1.34× on macOS**; **five benchmarks at ≥2× on both hosts**
— the whole document-scale marshal suite: ViperWrite **3.08×/3.04×**
(33 allocs/op → 1), Marshal/HugoFrontMatter **2.95×/3.00×**,
Marshal/ReferenceFile/map **2.92×/3.00×**, Marshal/ReferenceFile/struct
**2.18×/2.07×**, Marshal/SimpleDocument/map **2.09×/2.30×** — with
citm_catalog (~1.9×) and canada (~1.4×) behind them. Zero significant
regressions on linux.

The branch is merged with current v2: the new `unstable.Marshaler` /
RawMessage encoding (#1084) is integrated over the optimized encoder (shape
routing rides the entry classes; `rawShape` lives on the entry because
options are shared plan pointers here; the static-class shortcut is gated on
the marshaler mode), with the upstream RawMessage test suite passing.

Known trade-offs (kept, called out for review): twitter −7% on macOS only
(large converted-JSON objects; flat on linux) and a few macOS rows at ±1–3%.

Full report with per-benchmark charts:
https://files.pelletier.dev/reports/2026-07-02-go-toml-perf.html

## What changed

0a. **Static encoder-plan facts** (`marshaler.go`): struct entries re-derived
   per encode what their plan already knows — whether the key needs quoting
   (a byte scan per key), the entry class (two resolve+property lookups per
   entry), and the type properties (interface-hashed cache lookups, 20% of
   struct marshaling). Plan fields now carry the statically-knowable facts.

0b. **Reflection-free generic marshal** (`marshaler.go`): encoding a
   `map[string]interface{}` document wrapped every access in a
   `reflect.Value` even though the concrete types are known. Entries
   collected from generic maps now carry plain interface values and the tree
   encodes through type switches, sharing all layout code; only table
   recursion and non-generic values return to the reflection encoder. This
   is the symmetric counterpart of the fused decoder and took the
   generic-marshal family 2–3×.
1. **Seen-tracker rewrite** (`internal/tracker/seen.go`): key lookups scanned
   every entry recorded after the parent — quadratic for many-key tables
   (`find` was 33% of citm_catalog). Entries now keep their named children in
   a sibling chain (zero hashing for small/medium tables), spilling a
   parent's children to a shared seeded open-addressing index past 64
   children. Array-table refresh is O(1) (fresh id instead of compaction);
   arrays of scalars create no tracker entries at all; and container-value
   internals — which no later expression can extend or redefine — validate
   against a separate per-value tracker instead of accumulating in the main
   one for the rest of the document.
2. **Exact wide floats**: 17-significant-digit values (coordinate data)
   overflow Clinger's 53-bit fast path and fell back to `strconv` plus a
   string allocation each. For `|exp10| ≤ 27`, `5^|q|` is an exact `uint64`,
   so the correctly rounded value is computed with exact 128-bit integer
   arithmetic (multiply, or divide keeping the remainder, plus explicit
   round-to-nearest-even). Verified bit-exact against strconv over millions
   of randomized + directed cases; a deterministic sweep is in the tests.
3. **Decimal digit scanning** specialized (indirect call per byte → inlined).
4. **Fused native containers** (generic targets): arrays and inline tables
   decode straight into native values with no AST. Keys declared inside a
   container are logged and replayed through the tracker after the line
   finishes parsing, so error precedence (syntax → outer key → inner
   duplicates in document order) is unchanged.
5. **Fused struct decoding** (`decode_struct_fused.go`): scalar key-values of
   reflection targets also decode without an AST, dispatched through the
   existing descend/assign machinery via path parts carrying the scanned
   bytes and ranges. Table headers and container values still go through the
   expression parser (a two-function cursor bridge), so array-table
   bookkeeping and error rendering are identical; strict mode reports
   through parts with the same key spans an AST node would produce.
6. **Encoder**: table entries classified once (was up to three times, each
   array classification resolving every element); kind-first dispatch before
   `reflect.Type` comparisons; word-at-a-time `canBeLiteral`/
   `appendBasicString`; plain-string map keys skip TextMarshaler checks; one
   reusable `MapIter` + a value slab for larger maps; MRU memos for type
   properties and struct plans.
7. **Strict mode**: the key tracker keeps raw part bytes and only
   materializes strings when an error is recorded (was one allocation per
   key part per table header on every valid document).

## Compatibility

- No API changes; no behavior changes intended. Full suite (incl. the
  generated toml-test corpus) and `-race` are green; error messages and
  precedence are kept identical by construction and covered by the existing
  error tests plus new ones for the fused paths.
- Package coverage 96.8% vs 97.0% base; the residual is defensive branches
  mirrored from the AST loop that the parser cannot currently reach.
- `./caps.sh check` fails identically on base and branch in my environment
  (capslock flag mismatch) — worth re-checking in CI.

## Follow-up candidates

- A compiled-setter fast path for struct plans was prototyped and **measured
  at ≤2% on one benchmark** (reflect field access dominates, not dispatch),
  so it was dropped rather than added as a parallel path to maintain.
- twitter/macOS −7% (flat on linux): the remaining delta is diffuse across
  the fused inline-table machinery on arm64; a scope-map replacement for the
  duplicate replay was prototyped and regressed large container values, so
  the per-value tracker stayed.
- Moving generic decoding substantially further requires an opt-in
  visitor/streaming API (output-shape change) — design decision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0189ijMLcPYnX8H6mrucoAgE


pelletier:
Update: removed all `unsafe` usage per project rule (the codebase never uses unsafe) — the slab allocator now keeps only the pure-Go part (chunked subslicing of small `[]interface{}` backing arrays). Also added since the first push: fused (AST-free) decoding of scalar key-values for reflection targets. Re-running the full interleaved A/B on both hosts; will update the description with final numbers.

https://claude.ai/code/session_0189ijMLcPYnX8H6mrucoAgE

pelletier:
CI is fully green now (lint, coverage report, all matrices, fuzzing, capabilities).

One finding from the coverage work, pre-existing and unrelated to this branch: with `EnableUnmarshalerInterface`, a raw capture whose path crosses an `interface{}` step — e.g. `struct{ T interface{} }` with `T` pre-set to `map[string]unstable.RawMessage{}` and a `[t.x]` table — fails resolution with `toml: internal error: capture target does not implement UnmarshalTOML` (`resolveCapture`'s interface branch resolves to a fresh generic map instead of the existing typed one). The branch was uncovered on `v2` too. Happy to file it as a separate issue.

https://claude.ai/code/session_0189ijMLcPYnX8H6mrucoAgE

pelletier:
Closing this PR in favor of a split into per-technique PRs, so that each optimization documents its own impact and can be reviewed (and reverted) in isolation. The combined branch bundled ~13 separable techniques across the decoder, the seen-tracker, and the encoder — too much surface for one review.

The replacement series (each PR carries its own interleaved A/B benchstat vs its exact base):

**Independent (base `v2`, any order):**
- #1096 — exact wide-significand float parsing
- #1097 — specialized decimal digit scanning
- #1098 — word-at-a-time string encoding
- #1099 — sized reads, O(1) array-count reset, plan memo

**Decode chain (stacked, merge in order):**
- #1100 — sibling chains + adaptive hash spill, O(1) array-table refresh
- #1101 — native container decode for generic targets
- #1102 — allocation-free strict-mode key tracking
- #1103 — AST-free scalar key-values for reflection targets

**Encoder chain (stacked, merge in order):**
- #1104 — classify table entries once
- #1105 — type-properties fast paths
- #1106 — static field facts in encoder plans
- #1107 — reusable map iterator + value slab
- #1108 — reflection-free generic marshal

Differences from this branch, beyond the split itself:

- Rebased onto current `v2`: integrated with `SetOmitEmptySuperTables` (#1091) — its super-table check now rides the entry classes — and with the RawMessage suite from #1084 (already merged here).
- The native fused containers now enforce the same nesting-depth bound as the expression parser: this branch predates #1092, and its native container recursion silently bypassed that DoS fix. The split PR restores parity (`maxFusedNesting`), covered by the existing regression test.
- The dropped follow-ups (compiled-setter fast path, scope-map replay) stay dropped, as discussed in the description.

Combined, the series reproduces this branch's results. The integration of all 13 branches (`perf/split-integration`) is green (tests, `-race`, lint, module coverage ≥ base at every stage), and interleaved A/B benchstat of `v2` vs the integration gives:

- **linux/amd64** (t2d spot, go1.26.4, n=10): sec/op geomean **−23.3%**, allocs/op geomean **−33.8%** — ViperWrite −66.7%, Marshal/ReferenceFile/map −64.6%, Marshal/HugoFrontMatter −60.8%, citm_catalog −45.6%, canada −19.9%, code −10.8%; every suite benchmark improved except Unmarshal/SimpleDocument/map (+1.1%).
- **darwin/arm64** (M1 Pro, go1.26.4, n=10): sec/op geomean **−25.3%** — Marshal suite −53…−66%, ViperWrite −67.3%, citm_catalog −47.8%, canada −27.9%; twitter +7.8% is the same known trade-off called out above.

Each split PR's body carries the incremental benchstat vs its exact base, so the contribution of every technique is now measured separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #424, opened 2020-07-18
Multiline string generates invalid TOML on valid Go strings


danneu:
**Describe the bug**

Since the lib does less escaping when generating multiline strings, it doesn't generate an equivalent string between multiline and singleline mode.

I would expect (1) that flipping multiline=true can't break my toml and (2) that any valid Go string should produce a valid toml string. As it currently stands, you have to know in advance that your strings are "safe" to use with multiline which seems like a bug. I would expect multiline mode to merely use `"""` instead of `"`, and to avoid escaping newlines.

**To Reproduce**

Demo: https://repl.it/@danneu/RegalFreeLicense#main.go

```go
package main

import (
  "fmt"
  "github.com/pelletier/go-toml"
)

type Message1 struct {
  Text string
}

type Message2 struct {
  Text string `multiline:"true"`
}

func main() {
  msg1 := Message1{"Hello\\World"}
  msg2 := Message2{"Hello\\World"}
  toml1, err := toml.Marshal(msg1)
  if err != nil {
    panic(err)
  }
  toml2, err := toml.Marshal(msg2)
  if err != nil {
    panic(err)
  }

  fmt.Printf("toml1: %s\n", toml1)
  fmt.Printf("toml2: %s\n", toml2)

  // Message1 round-trips successfully 
  msg1parsed := Message1{}
  err = toml.Unmarshal(toml1, &msg1parsed)
  if err != nil {
    panic(err)
  }
  fmt.Printf("round tripped msg1: %#v\n", msg1parsed)
  // => round tripped msg1: main.Message1{Text:"Hello\\World"}

  // Message2 does not round-trip due to missing escape 
  msg2parsed := Message2{}
  err = toml.Unmarshal(toml2, &msg2parsed)
  if err != nil {
    panic(err)
  }
  fmt.Printf("round tripped msg2: %#v\n", msg2parsed)
  // => panic: (2, 1): invalid escape sequence: \W
}
```

**Expected behavior**

I would expect the same input to produce byte-equiv output and for Message2 to not break on valid Go strings.

**Versions**
 - go-toml: v1.8.0
 - go: 1.14
 - operating system: macOS

**Additional context**

This is because generateMultilineTomlString() does less escaping than generateTomlString() when I think they should do the same escaping with the one exception being \r and \n.

Maybe the current code was written at a time when you were using `'''` instead of `"""`?

AllenX2018:
Thanks for reporting.

danneu:
[...]

PR #1111, opened 2026-07-17
fix: validate temporal component ranges when marshaling

gaoflow:
Marshal formats the four temporal kinds (`LocalTime`, `LocalDate`, `LocalDateTime`, `time.Time`) through fixed-width formatters that never check their component ranges. Since the parser was tightened for the TOML 1.1 range rules (e.g. seconds must be `<= 59`), Marshal can now emit documents its own `Unmarshal` rejects, and in two cases emit output that silently decodes to a *different* value.

Reparse fails:

```go
b, _ := toml.Marshal(map[string]any{"t": toml.LocalTime{Second: 60}})
// b == "t = 00:00:60\n"
err := toml.Unmarshal(b, &map[string]any{})
// err: toml: seconds cannot be greater than 59
```

Reparses, but to the wrong value:

```go
// Nanosecond >= 1e9 rolls into the fraction: 1s becomes 0.1s
toml.Marshal(map[string]any{"t": toml.LocalTime{Nanosecond: 1_000_000_000}})
// -> "t = 00:00:00.100000000\n"

// An offset that is not a whole minute is truncated by RFC 3339 HH:MM:
// +01:01:01 loses its trailing second; a 30s offset collapses to +00:00 (UTC).
toml.Marshal(map[string]any{"t": time.Date(2024, 1, 1, 0, 0, 0, 0, time.FixedZone("", 30))})
// -> "t = 2024-01-01T00:00:00+00:00\n"
```

A `Precision > 9` additionally panics with `slice bounds out of range` in `LocalTime.String`.

This is the same class as the string (#424) and comment (#1081) fixes: Marshal should not produce TOML that Unmarshal refuses.

Fix: validate the components at the encoder leaf (`appendValue`), mirroring the ranges the decoder enforces, and return a descriptive error instead of emitting invalid or lossy TOML. `String()` is left untouched (it is also used for display). The audit covered all four kinds and every reachable component: hour/minute/second/nanosecond/precision, year/month/day (real calendar dates), and `time.Time` year plus zone-offset magnitude and minute alignment.

Behavior change worth flagging per CONTRIBUTING's compatibility note: these `Marshal` calls previously succeeded (with invalid, lossy, or in one case panicking output) and now return an error, matching `encoding/json`'s `UnsupportedValueError` convention. Values that were already valid are unaffected.

Tests are table-driven over every shape above (encode -> expect error, the two silent-corruption rows pinned), with green boundary controls (leap day, `23:59:59.999999999`, `+01:01`, year `0000`/`9999`, a `23:59` offset) and a Marshal -> Unmarshal -> Marshal round-trip over the valid boundary set.

`go test -race ./...` passes, coverage is unchanged (97.3%, new code fully covered), golangci-lint (pinned version) is clean, `./caps.sh check` reports no new capabilities, and benchstat on `BenchmarkMarshal` shows no regression.

### DIFF
--- a/marshaler.go
+++ b/marshaler.go
@@ -1452,6 +1452,66 @@ func isUnquotedKeyByte(c byte) bool {
 	return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '-' || c == '_'
 }
 
+// The temporal formatters below (String and time.Format) write their fields
+// blindly, so an out-of-range component would produce a document the parser
+// rejects, or, for a nanosecond >= 1e9 or a zone offset that is not a whole
+// number of minutes, one that silently decodes to a different value. Validate
+// against the ranges the decoder enforces and error instead of emitting
+// invalid or lossy TOML.
+
+func validateLocalDate(d LocalDate) error {
+	if d.Year < 0 || d.Year > 9999 {
+		return fmt.Errorf("toml: cannot encode local date with year %d out of range [0,9999]", d.Year)
+	}
+	if d.Month < 1 || d.Month > 12 {
+		return fmt.Errorf("toml: cannot encode local date with month %d out of range [1,12]", d.Month)
+	}
+	if maxDay := daysIn(d.Month, d.Year); d.Day < 1 || d.Day > maxDay {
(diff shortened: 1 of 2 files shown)


pelletier:
I'd argue that's a bug since it would be out of compliance with 1.1. Thank you!

PR #1116, opened 2026-07-23
fix: Encode with nil writer returns error

sonnemusk:
```go
toml.NewEncoder(nil).Encode(map[string]int{"a": 1}) // panic
```

Return `toml: Encoder writer is nil`.


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #1118, opened 2026-07-23
fix: strip UTF-8 BOM before unmarshaling

### LINKED ISSUE #250: Support UTF-8 with BOM
Now the lib doesn't support toml files in UTF-8 encoding with BOM

The lib correctly handles UTF-8 encoding with BOM files

sonnemusk:
## Summary

UTF-8 BOM (`EF BB BF`) at the start of a TOML document made `Unmarshal` / `Decode` fail with an invalid character on the first key. Common when files are saved from Windows editors.

Strip the BOM once at the start of decoding so both `Unmarshal` and `Decoder.Decode` accept these files.

Fixes #250.

## Test plan

- [x] `go test -run TestUnmarshalUTF8BOM`
- [x] Existing package tests

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -26,3 +26,11 @@ func TestParseDateTimeMissingTimezone(t *testing.T) {
 	_, err := parseDateTime([]byte("2021-01-01T00:00:00"))
 	assert.Error(t, err)
 }
+
+func TestUnmarshalUTF8BOM(t *testing.T) {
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing this PR as low-signal noise after a quality pass on open contributions. Sorry for the distraction — focusing on fewer, higher-impact fixes going forward.

